PR #5, opened 2018-02-23
Add contributors section

nipunn1313:
How does this look?


gvanrossum:
Since this is a Dropbox project maybe the Dropboxers don't need to be listed? We get paid to do this. Up to you -- if you feel you don't get to do this in "work time" I am fine with the current version.

Issue #4, opened 2018-02-21
Import for standard protobuf formats is wrong


nmiculinic:
Generates:

```python 
from google.protobuf.timestamp import (
    Timestamp,
)
```

and it should be:
```python
from google.protobuf.timestamp_pb2 import (
    Timestamp,
)
```

similarly for other standard types:
```python
from google.protobuf.empty_pb2 import (
    Empty,
)
```

gvanrossum:
Is that with the Go or the Python toolchain?

Can you easily fix the problem? If so, I recommend submitting a PR with the fix!

nmiculinic:
Python tooling.. I think it's an easy fix. Will do when I find some time, in a few days hopefully. For now I've fixed internally with sed (there are other places where standard python generation has some issues)

gvanrossum:
Thanks for your help!

nipunn1313:
Thanks @nmiculinic

PR #6, opened 2018-02-25
Fixes issue #4

nmiculinic:
This pull request fixes issue #4 

I've tested in on my own machine with protos which import standard google protobuffs, and custom protobuf implementation. python_out adds _pb2 to module name and should be added in imports.


gvanrossum:
@nipunn1313 is on vacation, he'll look at this when he's back in 1-2 weeks.

nipunn1313:
This looks great

(nothing wrong with a quick code review from the beach, right?)

PR #16, opened 2018-05-16
Use explicit optional in generated type stubs

smessmer:
This allows users to disable implicit optional when running mypy with the generated type stubs.


python/protoc-gen-mypy
```diff
@@ -160,11 +160,12 @@ class PkgWriter(object):
                     for field in [f for f in desc.field if f.label == d.FieldDescriptorProto.LABEL_REQUIRED]:
                         l("{} : {},", field.name, self.python_type(field))
                     for field in [f for f in desc.field if f.label != d.FieldDescriptorProto.LABEL_REQUIRED]:
+                        self._import("typing", "Optional")
```
dzbarsky: can you please do this as `optional = self._import("typing", "Optional")` and use `optional` var below? This matters because if there is an `Optional` message in the proto the import will be aliased under a different name. thanks!

smessmer:
fixed

nipunn1313:
Hi - can you paste an example before/after generated stub for this change into the commit message?

smessmer:
This is a simple enough change. It just adds `Optional[]` around the type of the fields, which should always be done when the type is defaulted to `None`.

PR #18, opened 2018-05-16
Use local imports to get other generated type stubs

smessmer:
When the protobuf type stubs don't live as top level packages, they need to import each other using local imports.

### DIFF
--- a/python/protoc-gen-mypy
+++ b/python/protoc-gen-mypy
@@ -83,7 +83,7 @@ class PkgWriter(object):
         for i, segment in enumerate(split):
             if segment and segment[0].isupper():
                 assert message_fd.name.endswith('.proto')
-                import_name = self._import(message_fd.name[:-6].replace('-', '_') + "_pb2", segment)
+                import_name = self._import("." + message_fd.name[:-6].replace('-', '_') + "_pb2", segment)
                 remains = ".".join(split[i + 1:])
                 if not remains:
                     return import_name


dzbarsky:
Thanks for the patch! Could you please provide an example proto and diff of the output with your change for ease of reviewing?

smessmer:
2 Proto files:
- https://github.com/onnx/onnx/blob/master/onnx/onnx.proto
- https://github.com/onnx/onnx/blob/master/onnx/onnx-operators.proto

Part of `onnx_operators_pb2.pyi` before this PR:

    [...]
    from onnx_pb2 import (
        NodeProto,
    )
    [...]

Same part of `onnx_operators_pb2.pyi` after this PR:

    [...]
    from .onnx_pb2 import (
        NodeProto,
    )
    [...]

Since both, `onnx_operators_pb2.pyi` and `onnx_pb2.pyi` live inside the `onnx` python package, this is important. Import with an absolute package name would need to specify `onnx.onnx_pb2`.

dzbarsky:
hmm to be honest i am not quite sure how relative/absolute import changed in py2/py3, will let @nipunn1313 / @gvanrossum approve this one when they're back

nipunn1313:
Thanks for the contribution! From the look of it this wouldn't work with nested imports, but those lack support anyway, so this should work.

Can you make sure to sign the CLA?
https://opensource.dropbox.com/cla

jcppkkk:
This PR cause mypy-protobuf generate error pyi files that fail mypy check.

For example, in https://developers.google.com/protocol-buffers/docs/reference/python-generated there is a example command:
```
protoc --proto_path=src --python_out=build/gen src/foo.proto src/bar/baz.proto
```
What if baz.proto imports foo.proto?
```
import "foo.proto";
```
It generate
```
from .foo_pb2 import (
```
and fails mypy check: `cannot find module named foo_pb2`; The same error happens to our project after update to 1.4. 

This change need to calculate relative filepath between current file to target file.

nipunn1313:
Thanks for the find @jcppkkk. I am going to look into testing (#22) to improve the quality of this project and increase confidence in contributions.

If I can get a fix-forward quickly, I'll attempt to do it, otherwise, we'll have to revert this change in the meantime.

I'll file a separate issue.

PR #27, opened 2018-06-02
Re-add py3 support now that typeshed supports

nipunn1313:


nipunn1313:
Hm, it's on the master of typeshed, but I believe we still need to wait for the master of mypy branch to pull forward its pin. Will keep this open for now and retry it when the pin pulls forward.

PR #36, opened 2018-06-27
Make _Stub classes concrete by reiterating all of the methods

msullivan:
Generated _Stub classes are intended to be concrete. Generate
non-abstract stubs for all of the methods in order to accomplish this.
This is kind of ugly, but so it goes.


nipunn1313:
lgtm thanks! This change makes sense.

PR #49, opened 2018-10-03
Gate generic service generation behind the file-descriptor option

### LINKED ISSUE #47: Gate the mypy generation of generic service stubs to only appear when py_generic_services is set.
We should gate the generation of generic service stubs in *_pb2.py to only appear when option py_generic_services = true; is set

See
https://developers.google.com/protocol-buffers/docs/reference/python-generated#service

Note that this is different than gRPC stubs which is covered in #46

nipunn1313:
Fixes #47


nickpresta:
Has this been released in any published version yet? I'm still seeing the previous (generate service stubs without the option being set) on version 1.7 which looks to be from September.

nipunn1313:
Not yet! I will do so later today! Good idea.
If you need it more urgently, there's really only one file
(protoc-gen-mypy) you need.

--Nipunn

nickpresta:
Thank you! I was more concerned that I missed the version or that it was being published elsewhere.

nipunn1313:
Should be published now

PR #52, opened 2018-10-12
Import message/enum types from other files even if they are not in a package.

### LINKED ISSUE #51: Import from other files si broken if the imported type is not in a package
Before importing a type, the current code checks whether it's in a package and then do not import it if it's not. I'll send a fix shortly.

pcorpet:
Fixes #51.


python/protoc-gen-mypy
```diff
@@ -69,30 +69,32 @@ class PkgWriter(object):
...
+                if segment and segment[0].isupper():
+                    assert message_fd.name.endswith('.proto')
+                    import_name = self._import(message_fd.name[:-6].replace('-', '_') + "_pb2", segment)
+                    remains = ".".join(split[i + 1:])
+                    if not remains:
+                        return import_name
+                    raise AssertionError("Don't support nested imports yet")
+                    # return new_nested_import(import_name, remains)
+
         if name[0] == '.' and name[1].isupper():
```
nipunn1313: The existing code comes in 3 phases
1) in file
2) in package
3) Elsewhere

It sounds like stage 1 was erroneously detecting package-less proto files. Is there a way we can keep these three stages (or potentially insert a stage 2.5 for packageless if that's easier), to keep this readable?

Can this goal be achieved with a simpler diff, that changes here:
`if message_fd != self.fd and name[0] == '.' and name[1].isupper()`
Would that be sufficient?

pcorpet: Actually, I think this logic is broken: in proto2 and proto3, Python generated files ignores package declaration and as such I believe that python type names should only depend on the files in which they are located. So it should be:
1. in file
2. elsewhere

PR #53, opened 2018-10-17
Add comments around packages

nipunn1313:
Additionally add @pcorpet to the contributors list.


nipunn1313:
Hi @pcorpet  - does this look good?
Would you like me to upload a version 1.8 to pypi which includes your latest changes around no-package imports?

pcorpet:
:+1:

No need for a version 1.8 on PyPi: I discovered that bug because of a bug in my own code (I had forgotten to add a package in some protos), and I've fixed my bug already.

PR #55, opened 2018-11-16
Give type of DESCRIPTOR class attribute for enums in the stub.

zozoens31:


nipunn1313:
Looks great! Thank you.

Issue #38, opened 2018-08-13
Support for nested imports


ikonst:
What will it take to add support for nested imports? This seems to be supported in the Go version but not in Python.

nipunn1313:
Hi! Thanks for the report!
Can you provide an example?

I'm out on vacation, but can hopefully take a look next week.

--Nipunn

nipunn1313:
Hey @ikonst  - I added some testing around this in #39, but my test example actually passed just fine.
Is there a practical situation in which you're running into the issue? It's possible my test didn't cover the extent of the issue. Anyway - I'm hoping that makes it easier to proceed here.

ikonst:
I managed to minimize it to:

- pb/my_message.proto:
  ```proto
  syntax = "proto3";
  package pb;
  message MyMessage {
      enum MyEnum {
          unknown = 0;
      }
  }
  ```

- pb/test.proto
  ```
  syntax = "proto3";
  package pb;
  import "pb/my_message.proto";
  message Test {
      pb.MyMessage.MyEnum my_enum = 1;
  }
  ```

ikonst:
I'd be interested to take a shot at this, since our main protobufs at Lyft are failing with this error. It might take me some time to get to it, though.

abhishekrb19:
@ikonst @nipunn1313 I'd be interested in this support too. Is this any progress or pointers on this? If not, I can take a shot at this.

ikonst:
@abhishekrb19, I didn't get a chance to work on this yet.

abhishekrb19:
@ikonst , I took a stab at it and was a relatively trivial change to make - please see https://github.com/dropbox/mypy-protobuf/pull/57

Let me know if you have any comments.

abhishekrb19:
Just an update - this was supported in https://github.com/dropbox/mypy-protobuf/pull/57.
Do you want to mark this issue resolved?

nipunn1313:
Yep thanks!

Fixed by #57

PR #57, opened 2018-11-25
Support for nested imports

abhishekrb19:
Addresses: https://github.com/dropbox/mypy-protobuf/issues/38

- Adds in nested imports support. This is really a one liner change since proto nested imports would translate to importing the corresponding package and referencing the inner classes in that package. 
- Adds test for protos and expected pyi for nested imports (enums and messages).

@nipunn1313 , please let me know if there is something else that needs to be added.


nipunn1313:
Hi! Thanks for the PR @abhishekrb19  - I will take a look

python/protoc-gen-mypy
```diff
@@ -91,9 +91,7 @@ class PkgWriter(object):
                 remains = ".".join(split[i + 1:])
                 if not remains:
                     return import_name
-                raise AssertionError("Don't support nested imports yet")
-                # return new_nested_import(import_name, remains)
-
+                return import_name + "." + remains
```
nipunn1313: Could you please leave a comment somewhere around here explaining what "remains" refers to - to make it clear to a peruser that it refers to a direct import of a nested enum or message from another package.

nipunn1313:
Hi @abhishekrb19 - this diff looks great! I added one small request around adding a comment for maintainability, but otherwise, I'm happy to approve!

Thanks for doing the nice test case - that really helped me review more easily.

Could you additionally take a look at Dropbox's CLA as well? Thanks.

python/protoc-gen-mypy
abhishekrb19: @nipunn1313 , done.

abhishekrb19:
@nipunn1313 , thanks for the review. I filled out the CLA and added a comment. 
Apologies for the delay.

nipunn1313:
Thanks!

PR #59, opened 2018-12-05
Strongly type the enums in the output of its own methods.

pcorpet:


nipunn1313:
I *think* there's an extra quotation in here, but otherwise it looks good

pcorpet:
I did the quotation on purpose as the type is not defined yet so Python won't parse it. Quoting it make it acceptable by python while still relevant for mypy.

nipunn1313:
When I tried locally without quotes (`s/{!r}/{}/`), python parses it just fine. I think when you are within the class, python has already parsed the definition (which is how recursive structures can exist).

Perhaps by adding a sample to `test/test_generated_mypy.py` and `test_negative/negative.py`, we can test out the behavior.

These files are run through mypy and at the end of the test run and are expected to succeed and fail respectively with the failures having some expected error messages. That'll be the only surefire way to defend against future regressions in the area.

Travis CI also checks this for you, so you can push it to CI to see if it's working.

pcorpet:
Indeed, it works. I've added samples in both tests to make sure there's no regression. Thanks for pushing!

pcorpet:
Side note, I'm not sure how I can use the Travis log: I don't have any error locally but there's a diff on Travis. Are the files uploaded as artifacts somewhere? or should we build a better error message showing the diff somehow?

pcorpet:
Yes, thanks I've rebased and it passed. However, I'm surprised that Travis compared it to the master branch. Does your travis test the PR code? or the code "of the PR if it was merged on latest master"?

Anyway, if this happens again, I guess that having the actual diff shown on Travis would help debugging faster.

Ready for another look.

PR #64, opened 2019-02-21
Type stringly-typed methods as literal types

### LINKED ISSUE #62: HasField, ClearField, WhichOneof are stringly typed - mypy doesn't catch issues
```def HasField(self, field_name: Text) -> bool: ...
def ClearField(self, field_name: Text) -> None: ...
def WhichOneof(self, oneof_group) -> Optional[Text]: ...
```

We could type these more strongly in mypy-protobuf using Literal, which would only allow valid field names.

Eg.
```
Union[Literal["field1"], Literal["field2"]]
```

Currently, if we try this in mypy-protobuf, we get errors of the format
```
dropbox/proto/nucleus/types_pb2.pyi:40: error: Argument 1 of "HasField" incompatible with supertype "Message"
```

however, I confirmed if we either
1) remove HasField from the supertype
2) mark the supertype as `Any` rather than Text

then mypy-protobuf can provide better typing on the specific instances of Message

I'll propose (1) to typeshed under the argument that you can't usefully call `HasField` on a Message without some additional knowledge of the real type.

This proposal would prevent a duck-typey use case
```
Message1 {
   a,
   b,
}
Message2 {
   b,
   c,
}
```

```
if m.HasField("b"):
    quack(b)
```

But if a caller is trying to ducktype, they could work around this by requiring `m` to be typed with a more specific protocol type.

For most cases, forcing HasField to be a literal will be better.

nipunn1313:
Fixes #62 

Started work on this
Still depends on
1) https://github.com/python/typeshed/pull/2807
2) mypy pulling forward its pin
3) Carefully mimicking behavior described in https://developers.google.com/protocol-buffers/docs/reference/python-generated#singular-fields-proto3
4) Additional testing

Currently this fails w/ a bunch of
```
test/proto/nopackage_pb2.pyi:46: error: Argument 1 of "HasField" incompatible with supertype "Message"
```


pcorpet:
How did you solve the errors you are mentioning?

I'm getting plenty of those since 1.10.

nipunn1313:
Need to use the latest mypy on master!
I actually realize that I jumped the gun on releasing 1.10 since mypy hasn't cut a release since the upstream to typeshed.

If you control the mypy revision you use, and simply use the latest master for the next few weeks, that should resolve it.

Anything after
https://github.com/python/mypy/commit/d87e626e2fbde93fe7958e62b38c07b23a4e7e32

which references this

https://github.com/python/typeshed/commit/07ea6614c8c03da598903d1fd4fcfd90a35a6ad3

PR #72, opened 2019-05-16
update filepath generation to match the behaviour of protoc --python-out

arussellsaw:
Hey! thanks for building this 🙏 

We're using some pretty long path proto defintions like:
github.com/monzo/foobar/service.my-cool-service/proto/mything.proto
and --python-out will replace all instances of `.` with a `/` for cleaner python imports or something.
in the python implementation mypy doesn't do this, so we end up with the _pb2.py files in a different directory to the _pb2.pyi files. this PR updates the generator to bring the two in line.

### DIFF
--- a/proto/test/proto/github.com/test/test.proto
+++ b/proto/test/proto/github.com/test/test.proto
@@ -1,0 +1,6 @@
+syntax = "proto3";
+package test;
+
+message TestMessage {
+	string foo = 1;
+}
--- a/python/protoc-gen-mypy
+++ b/python/protoc-gen-mypy
@@ -378,7 +378,8 @@ def generate_mypy_stubs(descriptors, response, quiet):
         assert name == fd.name
         assert fd.name.endswith('.proto')
         output = response.file.add()
-        output.name = fd.name[:-6].replace('-', '_') + '_pb2.pyi'
+        output.name = fd.name[:-6].replace('-', '_')
+        output.name = output.name.replace('.', '/') + '_pb2.pyi'
         output.content = HEADER + pkg_writer.write()
         if not quiet:
             print("Writing mypy to", output.name, file=sys.stderr)


arussellsaw:
also i don't seem to be able to run `sudo pip3 install git+https://github.com/dropbox/mypy-protobuf` is the version in pip different to this repo?

nipunn1313:
Hi! The diff looks ok!
Can you add a proto to proto/ which exercises this path - so we can see it in action in tests? Thanks. Adding the proto, then running `run_test.sh` should get it working.

```
pip3 install git+https://github.com/dropbox/mypy-protobuf#subdirectory=python
```

should work. It's because the root of the repository doesn't contain the setup.py.
I was seeing this error message which gave it away:
```
  Running command git clone -q https://github.com/dropbox/mypy-protobuf/ /private/var/folders/7b/mxbk8jt14tz60t48z7mff8y4xflnf9/T/pip-req-build-FlGGrb
    ERROR: Complete output from command python setup.py egg_info:
    ERROR: Traceback (most recent call last):
      File "<string>", line 1, in <module>
    IOError: [Errno 2] No such file or directory: '/private/var/folders/7b/mxbk8jt14tz60t48z7mff8y4xflnf9/T/pip-req-build-FlGGrb/setup.py'
```

See for details: https://stackoverflow.com/questions/13566200/how-can-i-install-from-a-git-subdirectory-with-pip

nipunn1313:
Hi @arussellsaw. Thanks for the contribution!
If you have a chance to get around to this, let me know.
Otherwise, with your blessing, I'll take it over when I get the chance.

zsluedem:
Any update on this one. I want it eagerly.

arussellsaw:
Hi both, sorry for the radio silence @nipunn1313 i had a quick go at adding some test protos but i think there are either failing tests in the repo or i'm doing something wrong, i'll commit what i have, but it'd be great if you could take over for me! 🙏

nipunn1313:
Closing in favor of #88 

Thank you so much for reporting this and coming up with the fix!

nipunn1313:
@arussellsaw @zsluedem - fix is up on pypi! Enjoy.

Issue #75, opened 2019-05-24
`run_test.sh` does not work out of a python3 virtual environment


nipunn1313:
Seeing
```
E           AssertionError: Pattern 'Unknown field garbage.' not found in 'Protocol message SimpleProto3 has no field garbage.'
```

PR #76, opened 2019-05-26
Bump to mypy 0.701 version

nipunn1313:


nipunn1313:
This fixes #75 

run_test.sh actually stopped working because the python version launched a new version somewhat recently that changed the error messages https://pypi.org/project/protobuf/#history

This diff pins to the old version (3.6.1) of python-protobuf in requirements.txt to avoid this problem in the future. I plan to try to bump the version of python-protobuf explicitly in a future diff.

This diff additionally bumps the mypy pointer to the latest release. Latest release no longer supports python3.4 which ships default in trusty (travis CI default)
As such, I bumped travis CI to explicitly use xenial so that we have a newer version of python3

[90% of software engineering is dependency management](https://en.wikipedia.org/wiki/Dependency_hell)..
Thank you @erictraut for reporting.

PR #78, opened 2019-05-27
Remove emission of python keywords.

### LINKED ISSUE #73: Syntax error output if a field is named 'from'
This tool generates .pyi files with syntax error when parsing a proto-file with a field named `from`. E.g. `foo.proto`:
```
message FooRequest {
	int64 from = 5;
}
```

Running mypy on a project with this kind of autogenerated files results in syntax error:
```
$ mypy
package/schemas/foo_pb2.pyi:356: error: invalid syntax
```

This is because the `foo_pb2.pyi` file has this in it:
```
class FooRequest(google___protobuf___message___Message):
    from = ... # type: int
```

The `from = ...` is invalid syntax, and so we have a problem.

Just to say I've set up my mypy to ignore these files, `mypy.ini`:
```
[mypy]

[mypy-package.schemas.*]
ignore_errors = True
```

But this doesn't ignore syntax errors. As per https://github.com/python/mypy/issues/6897 mypy **can't** ignore syntax errors. It would be favorable if this project did not generate syntax-invalid output.

I'm using versions:
```
$ python --version Python 3.6.7
mypy==0.701
mypy-protobuf==1.10
```

And here is a gist with the entire contents of the `.proto` file, the generated `.py` file (which does not have syntax errors and the problematic and unexcludable `.pyi` file:
https://gist.github.com/gaggle/8aef6102d314936c9eb33724194682b0

nipunn1313:
Added testing for this behavior.
Fixes #73. See #73 for details.


python/protoc-gen-mypy
```diff
@@ -40,6 +40,42 @@ else:
 GENERATED = "@ge" + "nerated"  # So phabricator doesn't think this file is generated
 HEADER = "# {} by generate_proto_mypy_stubs.py.  Do not edit!\n".format(GENERATED)
 
+# See https://github.com/dropbox/mypy-protobuf/issues/73 for details
+PYTHON_RESERVED = set([
```
gvanrossum: Might want to add `async` and `await` for Python 3.

nipunn1313:
Great catch! Added that in a follow up commit.

python/protoc-gen-mypy
ikonst: Can `keyword.iskeyword` be used for this?

nipunn1313:
Oh neat - yeah that would probably work.

--Nipunn

gaggle:
Love the fast turn-around on this, thank you! Just so I can set my expectations, when might a new release be cut? I couldn't find mention of when and how that's done in the repo, and currently latest release in [pypi](https://pypi.org/project/mypy-protobuf/#history) is 1.10.

nipunn1313:
Hi - I'll look into cutting one relatively soon (hopefully by end of month). Just waiting on one more diff to get in.
Thanks for using and contributing to mypy-protobuf!

PR #79, opened 2019-05-27
Add enum literals support

### LINKED ISSUE #56: Mypy errors out on enum and nested types (they are mis-annotated?)
I noticed `mypy` check fails on enum and nested types usage. I tweaked the `test_generated_mypy.py` to minimally reproduce the issue. 

[https://github.com/dropbox/mypy-protobuf/blob/master/test/test_generated_mypy.py#L52](https://github.com/dropbox/mypy-protobuf/blob/master/test/test_generated_mypy.py#L52
)

Passing an enum instead:

`s = Simple1(a_enum=1)
`

The example runs fine though, but `mypy` complains of incompatible type. Please see error below:

```
mypy --py2 test/test_generated_mypy.py
```
> test/test_generated_mypy.py:70: error: Argument "a_enum" to "Simple1" has incompatible type "int"; expected "Optional[OuterEnum]"

Please see the type hint here:
https://github.com/dropbox/mypy-protobuf/blob/master/test/proto/test_pb2.pyi.expected#L126

Should the type hint say `int` (or `str`?) instead of the enum type (`OuterEnum` here)? Or am I missing something here?

nipunn1313:
Fixes #56


ikonst:
Could you provide some samples illustrating what changes?

nipunn1313:
There are some samples in the changed files (see the .pyi file changes)

ikonst:
Given this:
```protobuf
enum MyEnum1 {
   ENUM1_FOO = 1
}

enum MyEnum2 {
   ENUM2_BAR = 1
}

message MyMessage {
    MyEnum1 enum1 = 1;
    MyEnum2 enum2 = 1;
}
```
we seem to now allow:
```python
message = MyMessage()
message.enum2 = ENUM1_FOO
```

nipunn1313:
yep - it's quite unfortunate though since it's valid underlying code, but clearly not what the author intended.

Perhaps we can work around this issue by using `NewType` to wrap the int value
https://mypy.readthedocs.io/en/latest/more_types.html#newtypes

NewType wrapper types are different to mypy, but the same at runtime.

nipunn1313:
Ok lets go with it!

Kami:
First thanks for adding enum literal support.

This indeed seems to fix direct enum literal notation, but it breaks the ``Value()`` method scenario which takes a string (enum name) and returns an enum value.

For example, with the following definition:

```protobuf
enum ExecutionStatus {
    REQUESTED = 0;
    SCHEDULED = 1;
    RUNNING = 2;
}
```

This used to work:

```bash
from messages_pb2 import ExecutionStatus

status_str = 'requested'
obj.status = ExecutionStatus.Value(status_str.upper())  # this should return enum value (aka Enum type / int)
```

Now it returns the following error:

```bash
error: Incompatible return value type (got "Union[Literal[0], Literal[1], Literal[2]]", expected "ExecutionStatus")
```

It may also brake ``Name()`` and other enum class methods, but I didn't have time to dig in yet.

nipunn1313:
In your example, what is obj and status? And yes, lets move this to a new regression issue.
Thanks

nipunn1313:
I poked around a bit and I believe that the issue here is that `obj.status` is typed to `ExecutionStatus`, but it would now need to be typed to `ExecutionStatus.ClosedValueType`

since ExecutionStatus is the enum itself and `ExecutionStatus.ClosedValueType` is the type of the Values in the enum (and `ExecutionStatus.ClosedKeyType` is the type of the Keys in the enum)

The merged PR was non-backward-compatible in that it changed the return value of Typing to literals, and move the int-y-ness of the enum to the `ClosedValueType` member.

This is more accurate based on this example I found
```
enum UserFSWAction {
   UNKNOWN = 0;
   ALLOW = 1;
}
```

```
>>> type(ALLOW)
<class 'int'>
>>> ALLOW
1
>>> UserFSWAction.Name(1)
'ALLOW'
>>> UserFSWAction.Value('ALLOW')
1
>>> type(UserFSWAction.Value('ALLOW'))
<class 'int'>
```

@Kami - does that make sense directionally?

I'll hold off on publishing to pypi until we have some alignment here. I'm ok with making a non-backward compatible change, but I'll want to have some clear changelist and bump the big version number.

Couple more considerations for proposal

change typing of the constants to be the ClosedValueType for better error messages
```
- ALLOW: typing_extensions___Literal[1]
+ ALLOW: UserFSWAction.ClosedValueType
```

change `ClosedValueType` to just `ValueType` to simplify notation required by callers.

nipunn1313:
for now I will revert the enum_literals change so we can push the version here, and I'll create a new task with the proposal.

PR #81, opened 2019-06-08
Support inner enum constants

jaens:
In Python protobufs, enum members can be accessed both as `module.ENUM_MEMBER` or `module.MyEnum.ENUM_MEMBER`.

This commit adds support for the latter style.

It also fixes a problem with using reserved keywords as an enum member name (declarations for those are skipped).

### DIFF
--- a/test/proto/nested/nested_pb2.pyi.expected
+++ b/test/proto/nested/nested_pb2.pyi.expected
@@ -56,6 +56,9 @@ class AnotherNested(google___protobuf___message___Message):
         def values(cls) -> typing___List[ClosedValueType]: ...
         @classmethod
         def items(cls) -> typing___List[typing___Tuple[ClosedKeyType, ClosedValueType]]: ...
+        INVALID: typing_extensions___Literal[0]
+        ONE: typing_extensions___Literal[1]
+        TWO: typing_extensions___Literal[2]
     INVALID: typing_extensions___Literal[0]
     ONE: typing_extensions___Literal[1]
     TWO: typing_extensions___Literal[2]
@@ -75,6 +78,9 @@ class AnotherNested(google___protobuf___message___Message):
             def values(cls) -> typing___List[ClosedValueType]: ...
             @classmethod
             def items(cls) -> typing___List[typing___Tuple[ClosedKeyType, ClosedValueType]]: ...
+            UNDEFINED: typing_extensions___Literal[0]
+            NESTED_ENUM1: typing_extensions___Literal[1]
+            NESTED_ENUM2: typing_extensions___Literal[2]
         UNDEFINED: typing_extensions___Literal[0]
         NESTED_ENUM1: typing_extensions___Literal[1]
         NESTED_ENUM2: typing_extensions___Literal[2]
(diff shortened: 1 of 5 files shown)


ikonst:
This is a new 3.8.0 feature:
https://github.com/protocolbuffers/protobuf/pull/5234

We should only generate such inner members when generating stub for 3.8-generated code.

jaens:
Sure, how should I check for that?

nipunn1313:
I'm ok with adding it universally since the change appears backwards
compatible to folks using older protobuf (for simplicity sake). I'm on
vacation - will take a look next week when I'm back! Thanks for the PR!

ikonst:
Backwards compatible = no false negatives?

jaens:
I assume backwards compatibility means that if you drop this change into *any* existing project, it will still properly typecheck (which it will).

test/proto/nested/nested_pb2.pyi.expected
```diff
@@ -54,9 +54,12 @@ class AnotherNested(google___protobuf___message___Message):
...
         def items(cls) -> typing___List[typing___Tuple[str, AnotherNested.NestedEnum]]: ...
-    INVALID = typing___cast(NestedEnum, 0)
-    ONE = typing___cast(NestedEnum, 1)
-    TWO = typing___cast(NestedEnum, 2)
+        INVALID = typing___cast(AnotherNested.NestedEnum, 0)
+        ONE = typing___cast(AnotherNested.NestedEnum, 1)
+        TWO = typing___cast(AnotherNested.NestedEnum, 2)
+    INVALID = typing___cast(AnotherNested.NestedEnum, 0)
+    ONE = typing___cast(AnotherNested.NestedEnum, 1)
+    TWO = typing___cast(AnotherNested.NestedEnum, 2)
```
nipunn1313: Would you mind adding a line to `test_func` in test/test_generated_mypy.py to cover this new usage?

nipunn1313:
This looks great! Unfortunately, it rebase conflicted against a recent change that landed (#79) but it should be relatively easy to resolve. After that, I can gladly approve. Thank you for the contribution!

test/proto/nested/nested_pb2.pyi.expected
jaens: Sure

jaens:
Rebased and added a type check test.

jaens:
Ah, looks like the tests need a protobuf version upgrade. Of course.

jaens:
@nipunn1313 Do you prefer requiring a newer version of Protobuf, or just changing the test so it does not actually execute the line (but still type checks it)?

Kami:
Thanks for contributing this PR.

I just reported the issue myself (https://github.com/dropbox/mypy-protobuf/issues/82) and wanted to start to work on the code changes and just now noticed the issue already exists :)

nipunn1313:
Go ahead and change the test to not execute the line w/ a comment
I'll look into requiring a newer version of protobuf later! (possibly w/ a test matrix w/ different versions)

jaens:
Alright, test updated

nipunn1313:
lgtm thanks so much!

One last thing! Can you sign the Dropbox Open Source CLA?

jaens:
sure

PR #87, opened 2019-07-15
Add checks of Python reserved words

atymoshchuk:
Added two more checks if  field's name is not one of Python keywords.


test/proto/test_pb2.pyi.expected
```diff
@@ -192,12 +192,6 @@ class PythonReservedKeywords(google___protobuf___message___Message):
     def FromString(cls, s: bytes) -> PythonReservedKeywords: ...
     def MergeFrom(self, other_msg: google___protobuf___message___Message) -> None: ...
     def CopyFrom(self, other_msg: google___protobuf___message___Message) -> None: ...
-    if sys.version_info >= (3,):
-        def HasField(self, field_name: typing_extensions___Literal[u"False",u"None",u"True",u"and",u"as",u"assert",u"break",u"def",u"del",u"elif",u"else",u"except",u"for",u"from",u"global",u"if",u"import",u"in",u"is",u"nonlocal",u"not",u"or",u"pass",u"raise",u"try",u"while",u"with",u"yield"]) -> bool: ...
```
nipunn1313: Eg - here. I think this is valid and correct mypy definition for HasField here.

nipunn1313:
Hi! Thank you for the pull request!

The main motivation to avoid emitting python keywords was to avoid generating invalid mypy. However, for HasField and ClearField, I believe that it is generating Literal arguments of strings. I believe the string "None" is considered valid mypy.

Is there a specific case in which generated mypy on python keywords for HasField and ClearField specifically lead to an issue?

Thanks!
-Nipunn

test/proto/test_pb2.pyi.expected
atymoshchuk: In previous mypy versions (I had 0.630) I had issues with "from" string, that's why had to add this check as well. With new mypy version on latest master everything works.

nipunn1313: Ah that's interesting. Well glad things are fixed up on latest mypy.
mypy-protobuf advertises support on 0.701 and greater, so hopefully this confusion won't affect others going forward.

nipunn1313:
Thanks for raising the issue! Hopefully this shouldn't be needed going forward. Please feel free to reopen if it continues to be an issue.

Issue #11, opened 2018-03-29
Add tests for go/python output equivalency


nipunn1313:
For existing places in which they are not equivalent, discuss/determine which behavior makes sense.

drather19:
Besides the concern raised in #12, I've noticed a few other differences:
- Differences in the package imports in the generated Python code
- Usage of bytes vs. str in the argument to FromString()

dzbarsky:
I believe `FromString` should take in `bytes` (the go version currently uses `str` because it was only used for python2 where they are equivalent). @nipunn1313 if you agree i can update it?

nipunn1313:
From https://developers.google.com/protocol-buffers/docs/reference/python-generated
```FromString(s): Returns a new message instance deserialized from the given string.```

So `s` is serialized, and should be `s: bytes`. I am in support.
Typically, in mypy for compatibility, we should strive to use either `typing.Text` or `bytes`, but never `str`, because `str` changed meaning across python versions.

nipunn1313:
planning on deprecating the go version instead!

nipunn1313:
dropped go support!

PR #89, opened 2019-07-20
Add typing for Message.DESCRIPTOR

shabbyrobe:
At the moment, there is typing information available for `MyEnum.DESCRIPTOR` but not for `MyMessage.DESCRIPTOR`.

I thought there might have been some non-obvious reason that it wasn't there to begin with -- I had a bit of a spelunk through the issues and history to see if there was a reason -- but came up empty handed and thought I'd just submit something anyway.

I accept the [Dropbox Individual Contributor CLA](https://opensource.dropbox.com/cla/individual/)

_Edit_: Just noticed there's a Go version in here too and the README states that Dropbox use it and would like it to have parity with the Python version. I've updated it, but I don't see any tests for it. What's the actual status of the Go version? The output seems a bit different. If it's unused, I'll leave it alone but if it's used and should be a first-class citizen, I can add a test or two.


nipunn1313:
This looks great! Would be awesome if you could update `test_generated_mypy.py` with a usage of this Message.DESCRIPTOR as that code is run by python as well as by mypy. A usage there would confirm that this field exists and runs in the protoc generated python + is also typed by mypy.

We have a larger issue #11 for go/python equivalency tests, but we don't have them yet. At Dropbox, we use both versions internally in different parts of the company, so we get a bit of trial-by-fire. I would say - don't worry about writing tests for the go part - we should just wait until #11 is resolved.

Overall LGTM thanks!
If you would like, you can also add yourself to the README as a contributor.

shabbyrobe:
Thanks for the super fast review!

I've added tests to both `test/test_generated_mypy.py` and `test_negative/negative.py`, not 100% sure if I've done it right but it seemed to fail when I did it wrong and work when I did it right!

Given the lack of tests, would you prefer I removed the updates to the Go version or should I leave them in? I did some testing locally to verify the output was valid, but I guess I'm just a bit nervous about breaking someone's workflow!

nipunn1313:
Looks great! Thanks.
We can leave the updates in. We'll fix em  if there's an issue.

shabbyrobe:
Wonderful! Thank you!! When do you expect the next release to go out?

nipunn1313:
I usually just kick one whenever there's a meaningful change. I'll go ahead and kick one out now!

PR #90, opened 2019-07-21
Add install_requires to python setup.py

### LINKED ISSUE #86: Add `protobuf` to setup.py
When using mypy-protobuf, the plugin expects `pypi/protobuf` to be installed.
If it's not, the build fails:
```
Traceback (most recent call last):
  File "/Users/gui_talarico/dev/repos/samara/by-schema/.venv/bin/protoc-gen-mypy", line 13, in <module>
    import google.protobuf.descriptor_pb2 as d
ModuleNotFoundError: No module named 'google'
--mypy_out: protoc-gen-mypy: Plugin failed with status code 1.
```

If `protobuf` pkg is required, seems to me we could just add it to `setup.py` so the dependency is automatically when `mypy-protobuf` is installed

Thoughts? 

```python
# setup.py
setup(
    install_requires=["protobuf"],
    # ...
)
```

nipunn1313:
Fixes #86

### DIFF
--- a/python/setup.py
+++ b/python/setup.py
@@ -11,6 +11,7 @@ setup(
     py_modules=[],
     url='https://github.com/dropbox/mypy-protobuf',
     download_url='https://github.com/dropbox/mypy-protobuf/archive/v1.13.tar.gz',
+    install_requires=['protobuf>=3.6.0'],
 
     scripts=['protoc-gen-mypy', 'protoc_gen_mypy.bat'],
 )


python/setup.py
```diff
@@ -11,6 +11,7 @@
     py_modules=[],
     url='https://github.com/dropbox/mypy-protobuf',
     download_url='https://github.com/dropbox/mypy-protobuf/archive/v1.13.tar.gz',
+    install_requires=['protobuf>=3.7.1'],
```
ikonst: Is this the minimal version allowed? In my organization we're still on protobuf 3.6 in some places.

nipunn1313: ah - I'm not sure. I know that 3.5 does not work, but did not try 3.6. This is the minimal version which is tested. (see https://github.com/dropbox/mypy-protobuf/blob/master/requirements.txt)

nipunn1313: 3.5 yielded an error at import time, so I couldn't test it
```
E   TypeError: __new__() got an unexpected keyword argument 'serialized_options'
```

3.6 has some small functional differences, but is probably mostly compatible in most cases
```
>       s.ClearField(u"a_string")
E       TypeError: field name must be a string
```

Seems like we may just want to require the latest tested version here, though your organization may be able to get away with a lower version - as long as you don't rely on the minor divergences we know of (unicode vs str on ClearField)
Note that this is not protoc version, but rather the python-protobuf version. (which are numbered similarly and can be confusing)

ikonst: Is mypy-protobuf calling `ClearField` with a `unicode` as part of its operation?

nipunn1313:
Nope - it just specifies the type of clear field to match the behavior in
3.7.1 (allows Unicode arguments)

I'm ok with loosening the dependency restriction to 3.6 here given that it
mostly works in most cases, but newer versions are still recommended.

--Nipunn

nipunn1313:
I went ahead and changed the requirement to 3.6.0 to be compat for you

However, we should note that as we only test against python-protobuf 3.7.1, it's a matter of time before we'll be relying on something that makes upgrade required. In the meantime, you will definitely have annotations for functions which may not yet have

PR #92, opened 2019-08-02
Alias builtin types 

reorx:
to prevent such case:

```
message Foo {
    google.protobuf.BoolValue bool = 1;
}
```

in the generated pyi file, the Foo class will look like:

```
class Foo:
    @Property
    def bool(self) -> google.protobuf.BoolValue: ...

    def HasField(self, field_name) -> bool: ...
```

but the bool in HasField definition has been rewritten as property method bool, which will result in mypy error as:
```
    error: Invalid type "foo_pb2.Foo.bool"
```

this commit add alias assignments for all builtin types that are used in the pyi, the generated results is like:

```
builtin___bool = bool
builtin___bytes = bytes
builtin___float = float
builtin___int = int
```

and the relevant type hints is like:

```
def HasField(self, field_name) -> builtin___bool: ...
```

thus the built-in variable collision could be prevented.


reorx:
Added an example of the case I described in `test3.proto`, you can reproduce the error by execute the following code on this branch:

```bash
$ git checkout master -- python/protoc-gen-mypy
$ bash run_test.sh
test/proto/test3_pb2.pyi:103: error: Invalid type "test.proto.test3_pb2.SimpleProto3.bool"
```

nipunn1313:
Great find! Thank you and thanks for adding a test case!

PR #93, opened 2019-08-05
add forward reference string literal support in enum

reorx:
Background:

Postponed Evaluation of Annotations https://www.python.org/dev/peps/pep-0563/

This PEP describes the support of referencing a type that is yet to defined in the annotation, and how to enabling it by calling `from __future__ import annotations` in Python after 3.7

In Python before 3.7 (<= 3.6), same thing can be achieved by adding single quote around the type, this is called [Forward references](https://www.python.org/dev/peps/pep-0484/#forward-references) in PEP 484.

---

Forward reference of type is only supported after Python 3.7 and with `from __future__ import annotations`, to make the pyi compatible with Python 3.6 (even if we assume it's for 3.7 and later, the pyi files are still incorrect because of missing the future annotations import), add foward_ref function to add single quotes around the enum type referenced inside the enum definition, whether single quotes is added is controlled by `FORWARD_REFERENCE_STRING_LITERAL` constant, could be a config/option in the future.

other changes:
- generate enum_full_name once before generate enum codes


python/protoc-gen-mypy
```diff
@@ -79,6 +79,17 @@ PYTHON_RESERVED = set([
     'yield'
 ])
 
+
+FORWARD_REFERENCE_STRING_LITERAL = True
```
nipunn1313: Hi! Can you add the spirit of the contents of your commit message as an inline message here explaining why this variable exists at all?
Thank you!

nipunn1313:
Additionally, I think you have to rebase after your previous PR landed!

python/protoc-gen-mypy
reorx: Sure. Because this PR aims at improving the compatibility and strictness of pyi file syntax, so I thought you guys may want to leave the choices to users who have different perspectives and requirements. Without this PR, the pyi files are still correct to mypy's rules (this is unlike my last PR), only not very friendly to IDEs (my PyCharm shows a lot of alerts for the enum part) or Python interpreter itself, I don't want to make the PR looks very arbitrary, but to provide the suggestions of my consideration.

Of course, if you think that using string literal for forward reference is Ok to be the default behavior, I'm glad to remove this variable for good.

reorx:
PR rebased, with an additional comment to add a local testing script.

nipunn1313:
lgtm thanks!

PR #94, opened 2019-10-01
Use absolute path to necessary python.

zifter:
It will prevent calling system python instead of python in venv.
Why system python was called instead of local? I don't know) But this fix helps.


nipunn1313:
Yep sounds good! Thanks for the contribution.

PR #98, opened 2019-11-14
Make it possible to install the tool just by running by go get

juzna:
This will make it easier to include the tool in Docker images, as the installation will be just one command in Dockerfile.


nipunn1313:
Yep! Chatted with @dzbarsky and this looks good. Thanks @juzna

PR #101, opened 2020-01-29
Make Python2 work with latests mypy + some fixes to tests for the latest mypy

mikolajz:


nipunn1313:
Great stuff @mikolajz - particularly around the testing! Thanks so much.

PR #102, opened 2020-02-05
Use entry_points:console_scripts to generate protoc-gen-mypy

chadrik:
This fixes an issue where the shebang line fails of the executable fails to work when the path to the python interpreter exceeds 127 characters on linux and 512 on darwin.

We ran into the problem on the Apache Beam project:  https://github.com/apache/beam/pull/10734

@udim chased down some info on this issue: https://github.com/pypa/virtualenv/issues/596#issuecomment-411485104

Along with a solution, as implemented in `distutils`: https://bitbucket.org/pypa/distlib/src/6818e71809c2779e5ad82e6017ea570eb590186f/distlib/scripts.py?at=master#lines-124

Below are the results of my own investigation into how to ensure that the `distutils` fix is applied.  I cloned mypy-protobuf and created a virtualenv with a path that exceeded the 512 character limit on darwin.

I tried installing the scripts in several ways:

| command | setup.py uses entry_points |  package format | produces #!/bin/sh |
| - | - | - |- |
| `pip install python/` | no |  wheel | no |
| `cd python/ && python setup.py install` | no | egg | no |
| `pip install python/` | yes |  wheel | yes |
| `cd python/ && python setup.py install` | yes | egg | no |
| `pip wheel -w . python/ && pip install mypy_protobuf-1.17-py2-none-any.whl` | yes |  wheel | yes |

This confirms that you need to use pip to install and entry_points in the project's setup.py file in order the fix to be applied to the installed script.  

All tests were performed with the following:

- python 2.7.16
- pip 20.0.2

IMHO, this change has the positive side effect of introducing a `mypy_protobuf` python library.


nipunn1313:
Whoa! This is awesome! Thanks @chadrik

chadrik:
My pleasure.  Thanks for the awesome project.  It would be awesome if you could get a new release of this out to PyPI soon, so that I can get my other merge request completed.  Thanks!

chadrik:
Hi @nipunn1313  what do you think about getting a new release out to pypi with this change?  It would be a huge help for me!

nipunn1313:
Thanks for the ping!
Just did it!

chadrik:
thanks!

PR #106, opened 2020-02-13
Even required proto2 fields can be omitted from constructor parameters.

mikolajz:


nipunn1313:
Thanks for the tests!

PR #120, opened 2020-05-13
Split Enum value type from EnumTypeWrapper

### LINKED ISSUE #118: Message enum attributes have incorrect type annotations
Running mypy-protobuf on the following proto definition
```
syntax = "proto3";

enum Enum {
  OK = 0;
  NOT_OK = 1;
}

message Message {
  Enum enum = 1;
}
```
Produces a .pyi file with
```
class Enum(builtin___int):
    DESCRIPTOR: google___protobuf___descriptor___EnumDescriptor = ...
    @classmethod
    def Name(cls, number: builtin___int) -> builtin___str: ...
    @classmethod
    def Value(cls, name: builtin___str) -> 'Enum': ...
    @classmethod
    def keys(cls) -> typing___List[builtin___str]: ...
    @classmethod
    def values(cls) -> typing___List['Enum']: ...
    @classmethod
    def items(cls) -> typing___List[typing___Tuple[builtin___str, 'Enum']]: ...
    OK = typing___cast('Enum', 0)
    NOT_OK = typing___cast('Enum', 1)
OK = typing___cast('Enum', 0)
NOT_OK = typing___cast('Enum', 1)
global___Enum = Enum

class Message(google___protobuf___message___Message):
    DESCRIPTOR: google___protobuf___descriptor___Descriptor = ...
    enum = ... # type: global___Enum
```
which suggests to mypy that the enum attributes exist on the enum object e.g. that  `Message().enum.OK` is valid. In reality the generated code (using grpc_tools) has `type(Message().enum) == int`, which obviously doesn't have the `OK` attributes (or `keys`, `values` etc) and `Message().enum.OK` fails at runtime

nipunn1313:
This fixes the conflation of the value type from the EnumTypeWrapper in the typing.

Added tests to confirm that we correctly error if used incorrectly.

Fixes #118

### DIFF
--- a/python/mypy_protobuf.py
+++ b/python/mypy_protobuf.py
@@ -205,10 +205,18 @@ class PkgWriter(object):
         # type: (Iterable[d.EnumDescriptorProto], Text) -> None
         l = self._write_line
         for enum in [e for e in enums if e.name not in PYTHON_RESERVED]:
-            # MyPy does not support referencing the enum by the unqualified name *inside* nested enums
-            enum_full_name = _forward_ref(prefix + enum.name)
+            enum_value_type = enum.name + "Value"
+            l(
+                "{} = {}('{}', {})",
+                enum_value_type,
+                self._import("typing", "NewType"),
+                enum_value_type,
+                self._builtin("int"),
+            )
+            l("{} = {}", _mangle_global(enum_value_type), enum_value_type)
+            enum_value_full_type = prefix + enum_value_type
 
-            l("class {}({}):", enum.name, self._builtin("int"))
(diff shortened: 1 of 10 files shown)


nipunn1313:
Note that this approach takes the opinionated approach of returning a NewType over the int

If you try to construct a message passing an int to the constructor, it will fail, expecting to be wrapped in the NewType

pcorpet:
FYI this is breaking our type checks, and it's not as simple as replacing MyEnum by MyEnumValue because MyEnumValue exists only during type checking. I'm converting all those types to string `MyEnum` => `'MyEnumValue'`

PR #125, opened 2020-05-15
Fix up enums to match with superclass

nipunn1313:
Depends on https://github.com/python/typeshed/pull/3995


dilipvamsi:
Hi @nipunn1313 ,

The errors in travis says that it is unable to find `_EnumTypeWrapper`
```python
from google.protobuf.internal.enum_type_wrapper import (
    _EnumTypeWrapper as google___protobuf___internal___enum_type_wrapper____EnumTypeWrapper,
)
```
I think it can be fixed by
```python
from google.protobuf.internal.enum_type_wrapper import (
    EnumTypeWrapper as google___protobuf___internal___enum_type_wrapper____EnumTypeWrapper,
)
```

nipunn1313:
travis postback seems to be flaky today
image

PR #128, opened 2020-06-05
Bump mypy version to 0.780

nipunn1313:


nipunn1313:
For some reason, travis is no longer writing back to the repo here. Looking into it.

Manually confirmed that the build kicked off and passed here:

image

Issue #32, opened 2018-06-02
When the protobuf type stubs don't live as top level packages, it's hard to use mypy-protobuf


nipunn1313:
Thanks to @smessmer for the report
Currently, mypy-protobuf uses absolute imports. If the protobuf stubs aren't a top level package, they won't be found.

Fixing mypy-protobuf would involve either
- Using relative imports everywhere (including in/out of nested packages)
- Prepending the "python_out" path from the protoc invocation (this is not provided to us through CodeGeneratorRequest, so it may need to be an additional argument)

Workarounds in the meantime might include
- rejigger the protoc invocation to set python_out="." (top level)
- Adding the sub path of the stubs to PYTHONPATH

nipunn1313:
Hiya. I'm going to close this because as far as I can tell, protobuf suffers from the same problem in python3
See https://github.com/protocolbuffers/protobuf/issues/1491#issuecomment-772720912

I'm not inclined to put work into this until protobuf itself supports python3 with the path to stubs being non-top-level

For now, if you want the stubs in non-root - the workaround of placing the path to stubs on PYTHONPATH is required for both the _pb2.py and _pb2.pyi files generated by protobuf and mypy-protobuf respectively

For what it's worth, the mypy-protobuf internal testsuite does this [link](https://github.com/dropbox/mypy-protobuf/blob/bd6bd64c8da7f3d8717a820e9b0aaa832b6b220e/run_test.sh)

PR #132, opened 2020-06-11
import other generated module relatively to package

Evgenus:
When proto files having dependencies, generated stubs imports each other in global fassion, which sometimes leads to broken imports. I've tried to fix that at least for simple cases.


Evgenus:
related to #32

nipunn1313:
This seems reasonable and would work. Have a couple of questions to make sure we're solving the issue

1 - what case leads to broken imports? It would be good to add some `.proto` files/package to the `proto` directory to illustrate this case to make sure we don't regress this in the future. Some test case `.proto` files that fail without your change and pass with your change will be great.

2 - I notice this diff covers relative imports within the same package. Does the same issue arise when importing something from a parent package, or another package within the same tree? Eg a relative import - eg `import ..parent.other_child_pb2`

Evgenus:
I'm glad you like it.

> 1 - what case leads to broken imports? 

Inside of generated files imports looks like `from folder.proto_file_pb2 import Message as ...` which works only for the case when the `folder is a global package`

> Some test case `.proto` files that fail without your change and pass with your change will be great.

Sure. Can you give me a hint of how to test if the import works inside of `.pyi`?

> 2 - Does the same issue arise when importing something from a parent package, or another package within the same tree?

I think yes. I will improve.

nipunn1313:
I believe if you add `.proto` files to the `proto/` directory, then those will autogenerate `.py` files via protoc and `.pyi` files via mypy-protobuf when you run the testsuite.

This may be tricky to test, since `test.proto` is a global level package in our instantiation. Perhaps to be fully generic, we don't actually want it to be a global level package in our testsuite.

The testsuite uses this line to run protoc - it does a dance to make sure that it generates absolute paths to packages. Perhaps there's something we can change on this line in order to illustrate the bug that your diff is fixing.
https://github.com/dropbox/mypy-protobuf/blob/master/run_test.sh#L20

Evgenus:
After all I realized that I don't actually need this. Accordingly to current implementation of protoc there is only one way of organizing python modules and it is based on imports from globaly available package. Also __init__.py need to be provided manually. So I will close that PR and just go with duplicating whole structure of packages from global module in folder with source .proto files.

nipunn1313:
yes - the __init__.py need to provided manually and it's quite bothersome =\
It's an issue w/ generating python as well

Issue #46, opened 2018-10-03
Plugin does not generate gRPC stubs to *_pb2_grpc.py


vmagamedov:
Starting from some version of the protoc compiler, gRPC service stubs are no longer generated into `*_pb2.py` files. Now they are generated into separate `*_pb2_grpc.py` files. So to properly support grpcio library `*_pb2_grpc.pyi` files should be generated for service stubs.

And it would be great to generate gRPC stubs optionally, this would be possible if two separate CLI options (plugins) for protoc were created:
- `--mypy_out` for messages
- `--mypy_gprc_out` for gRPC stubs

I'm asking for this feature because I'm the author of the [grpclib](https://github.com/vmagamedov/grpclib) library, which is an alternative gRPC implementation for asyncio, and users of this library don't need gRPC stubs for grpcio.

nipunn1313:
The plugin actually generates generic_services stubs (see here https://developers.google.com/protocol-buffers/docs/reference/python-generated#service)

mypy-protobuf currently hasn't been tested to work with GRPC stubs.

An additional flag `--mypy_grpc_out` sounds like a great feature.

nipunn1313:
I think the work item here is to create a second plugin alongside `protoc-gen-mypy` that looks like `protoc-gen-mypy_grpc` which generates gRPC stubs.

Additionally, we should gate the generation of generic service stubs in *_pb2.py to only appear when `option py_generic_services = true;` is set

vmagamedov:
Thank you for making service stub type hints generation optional. This totally suits me.

Wanted to note, that [protobuf](https://developers.google.com/protocol-buffers/docs/reference/python-generated#service) documentation also states that:
> If neither of the above lines are given, the option defaults to false, as generic services are **deprecated**. (Note that prior to 2.4.0, the option defaults to true)

So this is a deprecated feature. And I think that currently it is better to consult gRPC docs for gRPC-related work: https://grpc.io/docs/reference/python/generated-code.html, because all gRPC-related work moved into different projects and into plugins to generate appropriate stubs. And maybe this means that it is worthless to support `py_generic_services` option at all.

nipunn1313:
To be clear, the existing support in mypy-protobuf is *not* for GRPC. It is for `py_generic_services` which is part of the regular `--python_out` plugin

[...]

PR #133, opened 2020-06-11
grpc services generation

Evgenus:
Work in progress...

related to #46


nipunn1313:
cool!
Add a test case `.proto` file, commit the `.pyi.expected` files and add cases to `test_generated_mypy.py` to make sure the types are right and we'll be rolling!

python/mypy_protobuf.py
```diff
@@ -648,32 +735,52 @@ def _add_messages(messages, prefix, _fd):
...
 
 
-def main():
-    # type: () -> None
-    # Read request message from stdin
-    if six.PY3:
-        data = sys.stdin.buffer.read()
-    else:
-        data = sys.stdin.read()
+def wrap_code_generator(func):
```
nipunn1313, line 738: look into contextlib.contextmanager - which makes this kind of thing ergonomic

```
@contextmanager
def wrap_code_generator():
    // do before code

    yield

    // do after code
```

python/setup.py
```diff
@@ -12,6 +12,11 @@
     url="https://github.com/dropbox/mypy-protobuf",
     download_url="https://github.com/dropbox/mypy-protobuf/archive/v1.21.tar.gz",
     install_requires=["protobuf>=3.6.0"],
-    entry_points={"console_scripts": ["protoc-gen-mypy = mypy_protobuf:main"]},
+    entry_points={
+        "console_scripts": [
+            "protoc-gen-mypy = mypy_protobuf:main",
+            "protoc-gen-mypy_grpc = mypy_protobuf:grpc",
```
nipunn1313, line 18: nice!

python/mypy_protobuf.py
Evgenus: thanks. you are right. will change

python/mypy_protobuf.py
```diff
@@ -528,6 +529,77 @@ def write_services(self, services):
...
+            l(
+                ") -> {}: ...", self._output_type(method),
+            )
+            l("")
+
+    def write_grpc_services(self, services):
+        # type: (Iterable[d.ServiceDescriptorProto]) -> None
+        l = self._write_line
+        l(
+            "from .{} import *",
```
Evgenus, line 579: here I'm making an assumption that `mypy_out` and `mypy_grpc_out` are the same. We need somehow to use messages inside of services. How to do this better?

nipunn1313: Good callout - I'm not sure there's much we can do. You can see in our protoc instantiation in `run_test.sh` that there's already an assumption that py and mypy go to the same location. I think it's ok to make this assumption for now (no need to block on solving this)

Open to any creative ideas to enforce this.

nadavwe:
hey @Evgenus, I'm really in need for this capability. :)
is there any way I can help?
the code looks somewhat minimal - would it work in its current state?

Evgenus:
> hey @Evgenus, I'm really in need for this capability. :)
> is there any way I can help?
> the code looks somewhat minimal - would it work in its current state?

Hi @nadavwe, sorry for late reply. Terribly busy with my dues. 

You can use this PR as is. We do in our development. To try just fork the project and merge my branch (other branch with imports normalisation may also be required) into your copy and install package directly from your repository. It will do the job.

If you want to speed up this becoming part of the master you can help me introducing tests for it.

Sorry for delay, everyone. Will get back to this when the life gets less crazy. 😝

nadavwe:
thanks, it works perfectly ^_^

Zahlii:
I think there is an issue when dealing with top-level .proto files. 

using the following code: `#!/bin/bash
python -m grpc_tools.protoc -I./ --python_out=. --grpc_python_out=. --mypy_out=. --mypy_grpc_out=. *.proto`

Produces

`line 552, in write_grpc_services
    self.fd.name.rsplit('/', 1)[1][:-6].replace("-", "_") + "_pb2",
IndexError: list index out of range
`

python/mypy_protobuf.py
```diff
@@ -528,6 +529,77 @@ def write_services(self, services):
...
+            self.fd.name.rsplit('/', 1)[1][:-6].replace("-", "_") + "_pb2",
```
Zahlii, line 580: Will probably not work if there is no / inside the file name, i.e. dealing with top-level proto files. Why not using something like os.path.basename?

Zahlii: Actually seems like using [-1] will solve this issue

nadavwe: I actually used `Path(self.fd.name).stem`.
this removes the whole cryptic indices.

Evgenus:
@nipunn1313 I finally added tests, but grpc-stups relies supports only python>=3.6. Any suggestion?

I have time to put all attention on this. I hope we can resolve that PR before you change structure of the project completely (I saw the branch with messes).

python/mypy_protobuf.py
Evgenus: @Zahlii Thanks. I will fix.

nipunn1313:
Hi,
Thanks for the updates here.

I filed #164 for dropping support for running mypy-protobuf itself in earlier versions of python - however, I think it's still relatively important that mypy-protobuf can successfully type older code.

When you mention GRPC stubs only supporting py >= 3.6. I assume this is a runtime constraint of the GRPC program, rather than a constraint on the mypy-protobuf runtime. mypy-protobuf's runtime shouldn't be relevant.

The `run_test.sh` testsuite runs mypy in both py2 and py3 modes - to validate that the generated types work with both python2 and python3.

One idea may be to restrict run_test - to only generate the grpc stubs in supported python versions.

python/mypy_protobuf.py
```diff
@@ -643,34 +733,54 @@ def _add_messages(messages, prefix, _fd):
...
+        # Serialise response message
+        output = response.SerializeToString()
+
+        # Write to stdout
+        if six.PY3:
+            sys.stdout.buffer.write(output)
+        else:
+            sys.stdout.write(output)
 
+    return wrapper
```
nipunn1313, line 770: Slight preference for using contextlib.contextmanager over the wrapped() - but not going to block the diff on it (I can fix it later). I always find it confusing to follow the control flow on wrapped functions like this.

run_test.sh
```diff
@@ -29,6 +29,7 @@ find generated -type f -not \( -name "*.expected" -or -name "__init__.py" \) -de
     fi
     $PROTOC --mypy_out=generated --proto_path=proto/ --experimental_allow_proto3_optional `find proto/ -name "*.proto"`
     $PROTOC --python_out=generated --proto_path=proto/ --experimental_allow_proto3_optional `find proto/testproto -name "*.proto"`
+    $PROTOC --mypy_grpc_out=generated --proto_path=proto/ --experimental_allow_proto3_optional `find proto/testproto/grpc -name "*.proto"`
```
nipunn1313, line 32: Idea would be to limit this by an if statement around the python version.

run_test.sh
```diff
@@ -43,6 +44,7 @@ find generated -type f -not \( -name "*.expected" -or -name "__init__.py" \) -de
     source $MYPY_VENV/bin/activate
     if [[ -z $SKIP_CLEAN ]] || [[ ! -e $MYPY_VENV ]]; then
         python3 -m pip install setuptools
+        python3 -m pip install git+https://github.com/Evgenus/grpc-stubs.git@76aab756c87716566835600a6683710264890950
```
nipunn1313, line 47: What does this add here? I took a quick peek, but it's non-obvious why this is necessary.

Evgenus:
> When you mention GRPC stubs only supporting py >= 3.6. I assume this is a runtime constraint of the GRPC program, rather than a constraint on the mypy-protobuf runtime. mypy-protobuf's runtime shouldn't be relevant.

Runtime does not require stubs. Only tests do.

> The `run_test.sh` testsuite runs mypy in both py2 and py3 modes - to validate that the generated types work with both python2 and python3.

Can I specify which files to check for each version?

> One idea may be to restrict run_test - to only generate the grpc stubs in supported python versions.

What about CI? How to check exact version? I see in `run_test.sh` two different cases for python 2.7 and 3.5

nipunn1313 (commented):
A negative test seems valuable here as well - so that we can see what some of the expected mypy error messages will look like.

run_test.sh
Evgenus: this line works fine for all versions

run_test.sh
Evgenus: This is my fork of grpc-stubs, I tries to fix it so it would work for python 2.7. But it also fails to install in CI because of requirement to python version.

Evgenus:
> A negative test seems valuable here as well - so that we can see what some of the expected mypy error messages will look like.

I don't mind to add some, but what actually negative tests should do in case of GRPC?

run_test.sh
Evgenus: I guess I need to add if statement below where mypy reads generated files. not sure how to do it

nipunn1313:
> > When you mention GRPC stubs only supporting py >= 3.6. I assume this is a runtime constraint of the GRPC program, rather than a constraint on the mypy-protobuf runtime. mypy-protobuf's runtime shouldn't be relevant.
> 
> Runtime does not require stubs. Only tests do.

What are you referring to by runtime and tests here? To be clear, there's
1) Running mypy-protobuf-grpc to generate _grpc.pyi stubs
2) Running mypy on the _grpc.pyi stubs in [py27, py35, py38, ...] mode
3) Running the grpc program with python
At least, with the regular stubs, the run_test.sh script does all three. Which parts require python >= 3.6?


> What about CI? How to check exact version? I see in `run_test.sh` two different cases for python 2.7 and 3.5

run_test.sh is tricky. We run (1) under many versions of python (see the `.github/workflows/main.yml`)
We run (2) under 2.7 and 3.5 currently.
We run (3) under the same version of python as (1) currently - but this seems like an incorrect constraint.

Perhaps one thing I can try - is to adjust the CI matrix to make explicit what versions of python are supported in each of the three steps. It's a little tricky to install more than one version of python in the CI system, but perhaps we can make it work via something like pyenv.

In any case - figuring out CI python versions shouldn't block you. I think you can wrap the generation of grpc stubs on an if-statement for python version

nipunn1313:
> > A negative test seems valuable here as well - so that we can see what some of the expected mypy error messages will look like.
> 
> I don't mind to add some, but what actually negative tests should do in case of GRPC?

Idea is to add a line to `test_negative.py` which would be the sort of error we want mypy to catch (eg calling the service method with a typo, or calling a service method with invalidly typed arguments). The testsuite will run mypy over test_negative and expect mypy errors to appear. Those mypy errors are saved into `test_negative/output.expected.*` and are validated to make sure they don't regress in the future (and appear in the code review!).

nipunn1313:
Equivalently, in `test_generated_mypy.py` - you can add positive tests (valid usages of the GRPC stubs)

In the future, I can try to refactor the testing matrix to make this stuff more clear in the future. Sorry that it's confusing today. Will try to improve it down the line.

nipunn1313:
Would also like to request if you could add a little bit of documentation (an example) to README.md, some information to CHANGELOG.md under "upcoming" and of course, adding yourself to the list of contributors (in README.md)

I'm currently waiting on mypy to release so that I can release the current master as 1.24 (it's imminent - I've been in touch with mypy folks).

After that I'll be enthusiastic about merging this into 1.25 assuming we can get positive+negative tests to make it maintainable.

Thank you for coming back around to this. It's appreciated. Happy holidays!

nipunn1313:
If you can get the tests working/passing now - I'd be happy to do the legwork of getting it into 1.25 - but I'd need the positive tests to make sure that valid programs run in python AND pass mypy, and I'd need negative tests to make sure that invalid program raise reasonable-looking mypy errors

Evgenus:
> > > When you mention GRPC stubs only supporting py >= 3.6. I assume this is a runtime constraint of the GRPC program, rather than a constraint on the mypy-protobuf runtime. mypy-protobuf's runtime shouldn't be relevant.
> > 
> > 
> > Runtime does not require stubs. Only tests do.
> 
> What are you referring to by runtime and tests here? To be clear, there's
> 
>     1. Running mypy-protobuf-grpc to generate _grpc.pyi stubs
> 
>     2. Running mypy on the _grpc.pyi stubs in [py27, py35, py38, ...] mode
> 
>     3. Running the grpc program with python
>        At least, with the regular stubs, the run_test.sh script does all three. Which parts require python >= 3.6?

The separate package `grpc-stubs` (https://pypi.org/project/grpc-stubs/) requires python>=3.6. That package adds annotations to grps itself.
For the cases:
1. grpc-stubs are not required
2. grpc-stubs are mandatory or we will have errors reported by mypy. grpc implementation is not annotated.
3. in that case annotations and all stubs are either optional or completely irrelevant.

> run_test.sh is tricky. We run (1) under many versions of python (see the `.github/workflows/main.yml`)
> We run (2) under 2.7 and 3.5 currently.
> We run (3) under the same version of python as (1) currently - but this seems like an incorrect constraint.

Not sure I understand that.
 
> Perhaps one thing I can try - is to adjust the CI matrix to make explicit what versions of python are supported in each of the three steps. It's a little tricky to install more than one version of python in the CI system, but perhaps we can make it work via something like pyenv.

You can add to matrix separate env variable to indicate if we need to generate/check grpc.

> In any case - figuring out CI python versions shouldn't block you. I think you can wrap the generation of grpc stubs on an if-statement for python version

Generation works fine. The problem right now is only in running tests. Ok, I will add a check

Evgenus:
> In any case - figuring out CI python versions shouldn't block you. I think you can wrap the generation of grpc stubs on an if-statement for python version

Oh, I've got this. finally )) If we don't generate then no problems with check. So indeed we need if statement only around generation

Evgenus:
> run_test.sh is tricky. We run (1) under many versions of python (see the .github/workflows/main.yml)
> We run (2) under 2.7 and 3.5 currently.
> We run (3) under the same version of python as (1) currently - but this seems like an incorrect constraint.

Ahah. So there is external version of python defined by the system or CI. 

Ok the problem is that mypy with --python-version=2.7 can't read `grpc-stubs` because of encoding. So yeah, we need either to fix `grpc-stubs` or to exclude generate files from check (and I can't put if statement based on python version because it is a version of external python).

nipunn1313:
Ah yeah - it needs
`# -*- coding: utf-8 -*-`

Per
https://www.python.org/dev/peps/pep-0263/

nipunn1313:
I can look into refactoring the CI environment so that each test run runs a specific test case - so it's easier to specifically include generation of GRPC in only the relevant plans.

Give me a few days and I'll see if I can get it working.

Evgenus:
> I can look into refactoring the CI environment so that each test run runs a specific test case - so it's easier to specifically include generation of GRPC in only the relevant plans.
> 
> Give me a few days and I'll see if I can get it working.

Don't worry. Looks like I've managed to fix it. You can check if my approach is acceptable for you.

Meanwhile I will work on adding some negative tests

nipunn1313:
Nice negative test!

Similar test in test_generated_mypy - to show proper code being typed properly (and running properly at runtime) will help make this maintainable (eg if grpc has a change of behavior in the future, tests will catch it)

Evgenus:
> Nice negative test!
> 
> Similar test in test_generated_mypy - to show proper code being typed properly (and running properly at runtime) will help make this maintainable (eg if grpc has a change of behavior in the future, tests will catch it)

this is WIP. take a look at `run_test.sh`

nipunn1313:
bam! Got it working - splitting up the python versions into different build plans #168 - it's much nicer and should allow us to add/drop support more easily in the future.

Take a look @Evgenus 

I believe it should make it much easier for you to select in/out when grpc is used - and it should eliminate the need for you to manually clean artifacts on certain python versions in run_test.sh

I definitely plan on landing it - but wanted to give you a chance to look at it to see if you'd prefer to build your changes on top of it vs the other way around.

It does require installing pyenv on your local machine for development, but hopefully that's not too big of a requirement.

Evgenus:
@nipunn1313 I reviewed. Tell me when you done, so I can merge it into my branch.

Evgenus:
@nipunn1313 you can create another shell script that will execute `run_test.sh` with different env variables emulating test matrix locally

Evgenus:
Ok. It is stable ATM. You can check it. I think `run_test.sh` is a mess but works.

TODO:
[] positive tests
[] README.md
[] CHANGELOG.md

run_test.sh
```diff
@@ -46,23 +48,37 @@ find generated -type f -not \( -name "*.expected" -or -name "__init__.py" \) -de
...
     fi
 
+    # preparing isolated folder for testing mypy per python version
+    rm -rf generated_3.5
+    cp -R generated generated_3.5
+
+    rm -rf generated_2.7
+    cp -R generated generated_2.7
+    find generated_2.7 -name '*grpc.pyi' -exec rm {} \;
+    find generated_2.7 -name '*grpc.pyi.expected' -exec rm {} \;
```
nipunn1313, line 70: Don't need this stuff any more I think?
W/ the CI refactor, each run of `run_test.sh` only covers a single $PY_VER_MYPY_TARGET - and therefore should only need one `generated` directory.

run_test.sh
nipunn1313: Maybe limit this line by
`if $PY_VER_MYPY_TARGET != 2.7` ?

Also noticing that the previous line (line 38) is also compiling grpc protos based on the structure you've used.
Perhaps grpc test protos should go separately (eg proto/testgrpc?)

nipunn1313:
Ok nice it's working!

You're right, run_test.sh is a mess. Left some comments in there with ideas

Idea for positive tests:
Add positive tests in a separate file `test_grpc_usage.py` and only run those tests in supported configurations.

run_test.sh
Evgenus: Ok, I will replace with condition. Not it is possible

Evgenus:
@nipunn1313 I can call it done.

Evgenus:
@nipunn1313 Hi, any chance you can review it soon?

python/mypy_protobuf.py
nipunn1313: Ideally we wouldn't need this line - but rather above `self._import_message` would bring in mypy_out

generated/testproto/grpc/dummy_pb2_grpc.pyi.expected
```diff
@@ -0,0 +1,64 @@
...
+    Channel as grpc___Channel,
+    ServicerContext as grpc___ServicerContext,
+)
+
+from typing import (
+    Generator as typing___Generator,
+)
+
+
+from .dummy_pb2 import *
```
nipunn1313, line 20: eg here - if self._import_message is working correctly, then it should bring in the dependencies as needed.
Clearly something more tricky is going on. Ok if this isn't solved immediately, but adding a comment to the codegen to improve this would be nice.

nipunn1313 (approved):
Hi - it's looking good.
I cannot merge it until after 0.800 of mypy lands. After that I'll release 1.24, and THEN after that it will be a candidate to land.

Unfortunately, I will have to squash, because there are 38 commits and the commit history is a mess.
Typically, I like to see one-commit-one-idea with each commit having an independent idea (see https://git-scm.com/docs/SubmittingPatches#separate-commits for the idea). I'm less of a stickler about the finer points, but I definitely don't want a 38 commit history of amendments and merge commits going into the permanent history of the repo. Prefer to have the permanent history as a reflection of the ideas in your commit, rather than the day-to-day development process.

Usually, I'd like to  a sequence of commits, rebased on top of master, with each commit having an independent idea. git rebase --interactive is a great git tool for this
```
git checkout master
git pull
git checkout mybranch
git rebase -i master
```
There are nice tutorials on it if you google around.
I understand this is often a big ask, so in lieu of that, I'll squash the commits (or try to rebase it myself)

I'll approve it, since it's looking good, but we need to wait until mypy 0.800 and mypy-protobuf 1.24 release before merging. See - #170

generated/testproto/grpc/dummy_pb2_grpc.pyi.expected
Evgenus: `self._import_message` is already used for all messages. As you see it does nothing for messages withing same proto file.

Evgenus:
> and THEN after that it will be a candidate to land.

Things are going more and more complicated with this PR. I have no more time budget to invest into this. I guess the only way for me is to use my fork from now.

nipunn1313:
Sure thing. Thanks for taking the time to make it available upstream. There's no need for you to sink additional time into this if you aren't interested. I think it's in good enough shape that I can merge it later, once 0.800 / 1.24 get released.

nipunn1313:
1.24 launched. Will look into merging this now.
I'll handle the merge conflicts! Don't worry about it!

Evgenus:
@nipunn1313 Great news. I took a look at conflicts. Most of them are completely effortless to fix. Thanks. Really looking forward to see that in next version.

nipunn1313:
Closed by #176

Thanks so much @Evgenus - enjoy!

PR #140, opened 2020-06-25
Add map support

### LINKED ISSUE #138: Map Field does not support get_or_create
A protobuf map field provides an API which is slightly different from the normal Python dict (or MutableMapping). Instead, it acts like a `defaultdict`, and allows assignment to undefined keys. It does not, however, allow direct assignment when the field maps to a nested submessage.

Because this behavior is slightly unexpected, when the mapping's value type is a submessage, the map field provides a `get_or_create` method, as an alternative to direct assignment to undefined keys. From https://developers.google.com/protocol-buffers/docs/reference/python-generated#undefined:

> As it may not be immediately obvious to anyone reading your code that m.message_map[10] alone, for example, may create a submessage, we also provide a get_or_create() method that does the same thing but whose name makes the possible message creation more explicit:
```py
# Equivalent to:
#   m.message_map[10]
# but more explicit that the statement might be creating a new
# empty message in the map.
m.message_map.get_or_create(10)
```

Running mypy with stubs created with `mypy-protobuf` against code using map field's `get_or_create` method results in an `attr-defined` error.

### Test Case

```proto
// example.proto
syntax = "proto3";

message Nested {
    string value = 1;
}

message Map {
    map<string, Nested> mapfield = 1;
}
```

Compile using `grpc_tools` and `mypy-protobuf`:

```
python -m grpc_tools.protoc -I=. --python_out=. --mypy_out=. example.proto
```

Inspecting the stub file, we find the type assigned to the `mapfield`:

```py
    @property
    def mapfield(self) -> typing___MutableMapping[typing___Text, type___Nested]: ...
```
This does not include the extra `get_or_create` method supported by protobuf's map field. An example python file using `get_or_create` illustrates the error:

```py
# main.py
from example_pb2 import Map, Nested

m = Map()
mapping_1 = m.mapfield.get_or_create("foo")  # mypy attr-defined error!

nested = m.mapfield["bar"]
nested.value = "bar"
```

#### Mypy output:
```
$ mypy --show-error-codes --pretty main.py
main.py:4: error: "MutableMapping[str, Nested]" has no attribute "get_or_create"  [attr-defined]
    mapping_1 = m.mapfield.get_or_create("foo")  # mypy attr-defined error!
                ^
Found 1 error in 1 file (checked 1 source file)
```

#### Expected output

No mypy errors

### Library versions
```
grpcio==1.30.0
grpcio-tools==1.30.0
mypy==0.782
mypy-extensions==0.4.3
mypy-protobuf==1.23
protobuf==3.12.2
six==1.15.0
typed-ast==1.4.1
typing-extensions==3.7.4.2
```

Python version is 3.8.

## EDIT

The `get_or_create` method is only added when the map field is a map to a nested submessage. The bug still exists, but the example test case originally mapped to a scalar, not a nested message type. I've updated the test case accordingly.

nipunn1313:
Depends on https://github.com/python/typeshed/pull/4265
Will need to wait for a new mypy release to pick this up.

Fixes #138


nipunn1313:
This won't pass tests until mypy merges python/typeshed#4265 and releases a new revision that supports it.

PR #144, opened 2020-07-30
rename `_EnumTypeWrapper` to `EnumTypeWrapper`

danielorbach:
The module `google.protobuf.internal.enum_type_wrapper` does not contain the generated object `_EnumTypeWrapper`.

According to the history of google's [protobuf repository](https://github.com/protocolbuffers/protobuf) (master/v4/v3/v2.6.1) the fixed name has been used for the past few years.
Hence I feel safe to make this a simple fix, without consideration in the protobuf compiler/module installed.

### DIFF
--- a/python/mypy_protobuf.py
+++ b/python/mypy_protobuf.py
@@ -230,7 +230,7 @@ class PkgWriter(object):
                 "class {}({}[{}]):",
                 "_" + enum.name,
                 self._import(
-                    "google.protobuf.internal.enum_type_wrapper", "_EnumTypeWrapper"
+                    "google.protobuf.internal.enum_type_wrapper", "EnumTypeWrapper"
                 ),
                 enum_value_full_type,
             )


nipunn1313:
This is actually referring to the mypy base class here https://github.com/python/typeshed/blob/master/third_party/2and3/google/protobuf/internal/enum_type_wrapper.pyi
Which is generic over value - allowing us to wrap the value with a NewType

I don't think we actually want to rename this here (as shown by the tests failing w/ this change).

Would be happy to accept any change to the comments in the generated code which would make this more clear. Also open to other ideas on how to type this.

danielorbach:
Hmm I see... I encountered this only when enabling mypy actually.
What do you think about a conditional rename? Something like `"_EnumTypeWrapper" if MYPY else "EnumTypeWrapper"`

nipunn1313:
Can you describe the problem you're running into so we can explore the solution space?

I'm not sure I follow the issue. The code you edited autogenerates into .pyi files which are only interpreted by MYPY

nipunn1313:
I believe the generated code is working as desired here w/ `_EnumTypeWrapper` referring to the generic-over-value mypy base class - so the value can be customized to a NewType by mypy-protobuf.

I believe, in this case we don't want to inherit from `EnumTypeWrapper`, but rather from `_EnumTypeWrapper` here.

Usages of `EnumTypeWrapper` will actually be resolving to mypy as essentially `_EnumTypeWrapper[int]` while the autogenerated classes from mypy-protobuf have the richer type of `_EnumTypeWrapper[MyEnumValue]` with `MyEnumValue` being a `NewType` around `int` - for stronger typing.

45deg:
FYI: I've encountered a similar issue `error: Module 'google.protobuf.internal.enum_type_wrapper' has no attribute '_EnumTypeWrapper'; maybe "EnumTypeWrapper"?`, but after reinstalling the latest version of mypy, this is resolved.

nipunn1313:
Aha! Great sleuthing @45deg.

This Enum fix came in 1.22 of mypy-protobuf (CHANGELOG - https://github.com/dropbox/mypy-protobuf/blob/master/CHANGELOG.md#122) - which required mypy 0.780 or later

Perhaps I can more explicitly add this to the requirements.txt
mypy-protobuf the library does not directly import from mypy itself, but its generated code must be run by a newer mypy. It's technically not a requirement of mypy-protobuf working (IE generating the pyi files) - but it is indirectly a requirement of the full experience working.

PR #145, opened 2020-09-26
Add support for proto2 extensions

EPronovost:
Adds support for message extensions.  Now if there is a message 
```
message Bar {
  extend Foo { optional Bar ext = 2020; }
  
  optional bool flag = 1;
}
```
then `Bar.ext` will no longer raise a mypy error.  Furthermore, we are able to correctly type extension messages, so for something like
```
foo = Foo()
bar = foo.Extensions[Bar.ext]
```
the inferred type for `bar` is `Bar`.


test/test_generated_mypy.py
```diff
@@ -310,6 +320,17 @@ def test_which_oneof_proto3():
         s_untyped.WhichOneof("garbage")
 
 
+def test_extensions_proto2():
+    # type: () -> None
+    s = Simple1()
+
+    e1 = s.Extensions[Extensions1.ext]  # type: Extensions1
```
nipunn1313, line 327: May want to consider mypy-typing the `Extensions` variable on all messages to be a
```
Mapping[FieldDescriptor, google.protobuf.Message]
```

I'd also appreciate some kind of `assert isinstance(Extensions1.ext, FieldDescriptor)` to confirm that we're mypying it  to the correct type.

test/test_generated_mypy.py
```diff
@@ -310,6 +320,17 @@ def test_which_oneof_proto3():
...
+    e1.ext1_string = "first extension"
+
+    e2 = s.Extensions[Extensions2.foo]  # Will be type cast as an Any.
```
nipunn1313, line 330: If `Extensions` is typed, then e2 can by typed as a Message here.
It'll probably require some manual casting by the caller to get the appropriate message type, but I don't really see a way around that. I think it's probably better to force the caller to do an explicit type cast from Message to the appropriate Message type, over silently returning Any here.

In any case, what you have so far is better than what we had before, so I'm ok to run with it.

nipunn1313 (changes requested):
This looks great! Couple of comments around the testing, then we can go with it.

Trying to think of how we can add a negative test here - to confirm that this mypy typing catches bugs - has better typing than Any. Perhaps in `test_negative.py` we can add something of the sort
```
x = 5
x = Bar.ext   # E
```

Where we'd expect a type error of a FieldDescriptor being assigned to an int.

nipunn1313:
Thank you so much for the contribution!
Must mention that Dropbox requires that you agree to the CLA before it can be accepted https://opensource.dropbox.com/cla/

Thank you!

nipunn1313:
Would also ask that you update CHANGELOG.md with the new change in the "Upcoming" section

test/test_generated_mypy.py
EPronovost: I figured out a way to improve type inference for extensions.  Now we type `e1` as `Extensions1` explicitly.  Doing so required overwriting the `Extensions` defined for `Message`s in the typeshed.
I also added such `assert` statements.

test/test_generated_mypy.py
EPronovost: I figured out a way around it using generics :).  Now `e2` is inferred to be an `Extensions2`.

EPronovost:
The PR got more complicated, but we now have correct type inference for extensions.

nipunn1313:
Hmm - looks like this diff has developed some additional complexity for typing the extensions dict - and potentially has some points up for debate. If you were to split the PR into uncontroversial part (w/o typing extensions dict) - we can move forward with that easily, so we can focus discussion on this more controversial complexity and find the right way to solve it.

I don't want to hold up this entire diff on this discussion - since it's clearly adding value!
Thank you so much for doing this work!

EPronovost:
@nipunn1313 I made a more minimal PR here: https://github.com/dropbox/mypy-protobuf/pull/149
What would you like to discuss about this PR?

nipunn1313 (commented):
Sorry Forgot to hit submit on the comments! Here they are.

python/mypy_protobuf.py
```diff
@@ -536,10 +558,57 @@ def python_type(self, field):
         assert field.type in mapping, "Unrecognized type: " + repr(field.type)
         return mapping[field.type]()
 
+    def write_typeshed(self):
```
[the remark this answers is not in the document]
EPronovost: Yeah I removed this in favor of a template file.

python/mypy_protobuf.py
```diff
@@ -258,6 +261,13 @@ def write_messages(self, messages, prefix):
                     "DESCRIPTOR: {} = ...",
                     self._import("google.protobuf.descriptor", "Descriptor"),
                 )
+                # Overwrite the definition of Extensions from the standard
+                # protobuf typeshed.
+                l(
+                    "Extensions: {}[{}] = ...  # type: ignore[assignment]",
```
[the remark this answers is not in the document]
EPronovost: Previously the `_ExtensionDict` we defined did not inherit from the `ExtensionDict` defined in the standard typeshed, so although the methods were similar they were different objects.  I've changed this to fix it (note that there are still override ignores for the methods we're overwriting).

generated/_typeshed_mypy_protobuf.pyi.expected
```diff
@@ -0,0 +1,25 @@
...
+_ContainerMessageT = typing___TypeVar('_ContainerMessageT', bound=google___protobuf___message___Message)
+_ExtenderMessageT = typing___TypeVar('_ExtenderMessageT', bound=google___protobuf___message___Message)
+
+class _ExtensionFieldDescriptor(google___protobuf___descriptor___FieldDescriptor, typing___Generic[_ContainerMessageT, _ExtenderMessageT]):
+    pass
+
+class _ExtensionDict(typing___Generic[_ContainerMessageT]):
+    def __getitem__(self, extension_handle: _ExtensionFieldDescriptor[_ContainerMessageT, _ExtenderMessageT]) -> _ExtenderMessageT: ...
+
+    def __setitem__(self, extension_handle: _ExtensionFieldDescriptor[_ContainerMessageT, _ExtenderMessageT], value: _ExtenderMessageT) -> None: ...
```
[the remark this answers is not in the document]
EPronovost: > 1 - Why does this need to be autogenerated? Seems like it potentially should belong either directly in typeshed itself, or inside the mypy_protobuf library as something to be imported

Yes I agree the autogeneration is not the best solution.  I can think of 3 possibilities:

* Generate a file as part of the protoc plugin (i.e. what we have currently).  I changed it to use a template file instead of having python code to write the file on the fly.  The benefit of this is that we can ensure these stubs will be present whenever `mypy` is run and don't add any new package requirements.

* As part of the `mypy_protobuf` library.  This simplifies the generated code and still ensures that the same version of the base classes used to generate stubs is used to evaluate them (assuming the user doesn't change library versions in between).  However, I think it also means that `mypy_protobuf` would need to be present whenever `mypy` is run, which is not currently a requirement.

* As part of the `typeshed`.  This addresses the issue of needing `mypy_protobuf` present when running `mypy`.  However, it means that we'd require the new `mypy` version to have these stubs be present.  It would also pose difficulties if we want to change anything about this file in the future.

There may be a way to do either of the latter two options in a "soft" way; i.e. fallback to the minimal extension support already in `master` if the respective dependency is not present.  I don't have a ton of experience publishing to the `typeshed` to know if/how they support API changes across package versions.

> 2 - Can you explain what value this is adding over a Dict[ContainerMessage, ExtenderMessage]

Two reasons.  `Message.Extensions` is not a dictionary, and does not support the various other methods associated with one (e.g. you can't do `foo.Extensions.values()`).  Additionally, I don't think using a static typing would allow for the polymorphism we currently have (i.e. `_ExtensionDict` is a class parametrized by `_ContainerMessageT` and `__{gs}estitem__` is a method of that class parametrized by `_ExtenderMessageT`).  The actual source code is [here](https://github.com/protocolbuffers/protobuf/blob/e667bf6eaaa2fb1ba2987c6538df81f88500d030/python/google/protobuf/internal/extension_dict.py#L63).  I'm not sure why the typeshed implementation does not include the other methods; I added all the public ones here.

python/setup.py
```diff
@@ -8,10 +8,11 @@
     license="Apache License 2.0",
     author="Nipunn Koorapati",
     author_email="nipunn@dropbox.com",
-    py_modules=["mypy_protobuf"],
+    packages=["mypy_protobuf"],
```
EPronovost, line 11: I don't have any experience making a PyPI package... if there's a better way to do this let me know.

generated/_typeshed_mypy_protobuf.pyi.expected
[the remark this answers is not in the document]
nipunn1313: Ah! We can totally add these stubs to typeshed - as long as the stubs match the source code.
We have lots of precedent doing that (much of the google/protobuf typeshed stubs were contributed by dropboxers / folks who work on mypy-protobuf)

Source: https://github.com/protocolbuffers/protobuf/blob/e667bf6eaaa2fb1ba2987c6538df81f88500d030/python/google/protobuf/internal/extension_dict.py#L63

Typeshed: https://github.com/python/typeshed/tree/master/third_party/2and3/google/protobuf/internal
https://github.com/python/typeshed/tree/master/third_party/2and3/google/protobuf/internal

You can also test your typeshed change + mypy-protobuf change within mypy-protobuf by doing
```
CUSTOM_TYPESHED_DIR=~/path/to/your/typeshed/fork ./run_test.sh
```

Though one gotcha - when developing you should run the typeshed tests first - because errors within typeshed get suppressed when you use CUSTOM_TYPESHED_DIR.

It does mean that we'll have to wait for another release of mypy before we launch the next version of mypy-protobuf, but I think that's ok! #140 is similarly waiting on a typeshed change (you can look at that as an example)

python/mypy_protobuf.py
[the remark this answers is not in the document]
nipunn1313: Seems like we may be able to avoid this by updating the standard typeshed to be more like what we want!

nipunn1313:
Here are some examples of updates to typeshed for sake of mypy-protobuf
https://github.com/python/typeshed/pull/4265/files
https://github.com/python/typeshed/pull/3996/files
https://github.com/python/typeshed/pull/3995/files

generated/_typeshed_mypy_protobuf.pyi.expected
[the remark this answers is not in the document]
EPronovost: Interesting... I didn't realize the typeshed stubs for protobuf came from this same group (thank you!).  Are there any further comments you have on the current code before moving `typeshed.pyi` to the official typeshed?

python/setup.py
EPronovost: Will revert back once the typeshed is updated.

EPronovost:
This will be blocked on https://github.com/python/typeshed/pull/4619

nipunn1313:
the typeshed change landed, and I landed a mypy->typeshed pin bump afterward!

I spoke with the mypy maintainers (they're also at Dropbox), and they mentioned that it may be until around November before there's a mypy release w/ your change.
Feel free to update https://github.com/dropbox/mypy-protobuf/blob/master/run_test.sh#L36 to a git revision to unblock your change. In November (or whenever), I'll change it back before releasing mypy-protobuf

From
```
python3 -m pip install git+https://github.com/python/mypy.git@v0.780
```
to
```
python3 -m pip install git+https://github.com/python/mypy.git@7273e9ab1664b59a74d9bd1d2361bbeb9864b7ab
```
referencing https://github.com/python/mypy/commit/7273e9ab1664b59a74d9bd1d2361bbeb9864b7ab

test/test_generated_mypy.py
```diff
@@ -351,18 +351,16 @@ def test_extensions_proto2():
...
-    # Currently any extension is typed as an Any.
-    e3 = s1.Extensions[Extensions1.ext]
-    e3 = 3.1415
+    del s1.Extensions[Extensions2.foo]
 
-    # Currently we don't check that the extension is being applied to the
-    # correct base model. This fails at runtime but passes mypy.
+    for x in s1.Extensions:
+        assert isinstance(x, FieldDescriptor)
+        y = s1.Extensions[x]  # y should be typed as an AnyMessage
```
nipunn1313, line 358: This comment might leave some confusion as to the why, I think more specifically, explaining what "x" should be typed as will help understand why "y" is an `Any` here.

Specifically, to add clarity that `y` is `Any` because `x` came out of an iterator, which can't know the value type.

nipunn1313:
This is awesome!!! Thank you so much for the contributions here.

I left one message about a comment in the tests. Going to go ahead and merge it - feel free to add another PR to update the comment if you're excited.

PR #149, opened 2020-10-04
Minimal implementation of message extensions

EPronovost:
A simpler version of https://github.com/dropbox/mypy-protobuf/pull/145

This PR adds support for proto2 message extensions.  For a message
```
message Bar {
    extend Foo { optional Bar ext = 2020; }
    
    optional bool flag = 1;
}
```
then accessing `Bar.ext` will no longer cause a mypy error.  This PR does not improve the typing of the `Extensions` mapping from [what is defined in typeshed](https://github.com/python/typeshed/blob/master/third_party/2and3/google/protobuf/message.pyi#L10-L12), so any extension is inferred to be an `Any`.  See https://github.com/dropbox/mypy-protobuf/pull/145 for a way to improve this.


nipunn1313:
Nice! Thanks!

PR #154, opened 2020-10-25
Tests for extension methods

EPronovost:
Adds tests to match https://github.com/python/typeshed/pull/4706
Will fail until that PR is merged and we can update the `mypy` SHA


test_negative/output.expected.2.7
```diff
@@ -29,26 +29,28 @@ test_negative/negative.py:86: error: Invalid index type "str" for "_ExtensionDic
...
-test_negative/negative.py:130: note:     def get(self, k: int) -> Optional[unicode]
-test_negative/negative.py:130: note:     <1 more non-matching overload not shown>
-test_negative/negative.py:131: error: No overload variant of "get" of "Mapping" matches argument type "str"
-test_negative/negative.py:131: note: Possible overload variant:
-test_negative/negative.py:131: note:     def get(self, k: int) -> Optional[OuterMessage3]
-test_negative/negative.py:131: note:     <1 more non-matching overload not shown>
-test_negative/negative.py:134: error: Incompatible types in assignment (expression has type "Optional[unicode]", variable has type "int")
-test_negative/negative.py:135: error: Incompatible types in assignment (expression has type "Optional[OuterMessage3]", variable has type "int")
-Found 44 errors in 1 file (checked 19 source files)
+test_negative/negative.py:90: error: Argument 1 to "HasExtension" of "Message" has incompatible type "_ExtensionFieldDescriptor[Simple2, SeparateFileExtension]"; expected "_ExtensionFieldDescriptor[Simple1, Any]"
```
nipunn1313, line 32: Would be very cool if the `expected` line had a union of the possible extensions, but that seems fairly difficult - as it involves the rendering of a message to include information from all generated code - so this seems good enough!

nipunn1313:
This lgtm!

nipunn1313:
Please update the CHANGELOG.md!

I just tried updating to version 0.790 - but it looks like the earlier extension typeshed changes didn't make it.
I'll try and find out from mypy folks how long that's going to take - if it may be a while, I'll try to release 1.24 w/o
the extension changes - and punt the extension changes to 1.25. If it's all coming soon - we should be able to roll it
into 1.24.

EPronovost:
Thanks.  As soon as the typeshed PR merges I'll make a mypy PR to bump the typeshed version.  Hopefully that happens quickly and mypy makes a release shortly thereafter.

EPronovost:
PR to bump `typeshed` version in `mypy` is here: https://github.com/python/mypy/pull/9659

EPronovost:
@nipunn1313 I tried updating the mypy SHA in `run_test.sh`, but that seems to also have picked up other changes in mypy for the 0.80 release (specifically the version is `mypy==0.800+dev.985a20d87eb3a516ff4457041a77026b4c6bd784`).
But when I do I get a bunch of errors, starting with 
```
generated/google/protobuf/unittest_lazy_dependencies_pb2.pyi:11: error: Cannot find implementation or library stub for module named 'google.protobuf.unittest_lazy_dependencies_custom_option_pb2'
[import]
    from google.protobuf.unittest_lazy_dependencies_custom_option_pb2 import (
    ^
```
I'm not sure if any module discovery code changed in mypy.  Do you have any ideas what could be going on?

nipunn1313:
Investigated - and I'm fairly sure it's because typeshed deleted stubs for older versions of google.protobuf - referenced by our copy-pasted google/ .proto files. I removed them here as they aren't needed to run the tests - as those .proto files ship with protoc (up to date)

Separately - it was probably unintended for upstream folks to break older versions of protobuf - I'll make a task with them.

nipunn1313:
#155 should fix your issue! Then you should be able to bump the pin.

EPronovost:
@nipunn1313 Are you thinking that the next `mypy-protobuf` version will target `mypy` version 0.80 to pick up these changes to `typeshed`?

nipunn1313:
That's my current plan for simplicity.
Picking up 0.79 will require some complex branching/tagging of mypy-protobuf in order to pick up just the features that rely on it - so it'll be easier on me to just wait for 0.80. I'm following up w/ mypy team to get a sense of timeline, but I have to imagine it'll be within November.

nipunn1313:
Thanks so much for driving this through - and for adding unit tests. That's really helpful in making sure it keeps working - and also definitely makes it easier to code review - since it includes examples of what kinds of errors you'll see.

nipunn1313:
I'm actually not 100% sure this has to do w/ the typeshed stubs going away - looked in a bit deeper and am not sure any more - doesn't seem like those particular stubs were ever there. Might have to do with how they're being resolved from within the repo's pyi files.

In any case - we still don't need the google .proto files as it ships with protoc - and those ones are more up to date anyway.

EPronovost:
Yes something definitely changed about module discovery... the number of files listed in the negative error message went from 19 to 20 (the extra file was `mypy/mypy.proto`).  I'm not as involved in `mypy` directory to know what recent changes have occurred there.

PR #155, opened 2020-10-31
Remove the google .proto files in the repo

nipunn1313:
They come shipped with protoc - so it's not necessary to have them here.
They go stale w/ protoc upgrades if we keep them here.


EPronovost:
Looks like the negative output has changed.

PR #167, opened 2020-12-28
Run github workflows from any branch

nipunn1313:


nipunn1313:
for convenience - when developing on side branches

PR #168, opened 2020-12-28
Split CI matrix based on python version

nipunn1313:
Want to test running
mypy protobuf: 3.8, 3.7, 3.6, 2.7 [looking to drop 2.7 here]
mypy target: 2.7, 3.5 [looking to expand in future]
unit tests: 2.7, 3.8  [Validating real runtime behavior of google.protobuf]

Chose 6 key matrix plans to cover these cases (instead of full 3x2x2 matrix)

In the future, hope to split out the mypy of mypy_protobuf.py itself
into a separate step - which only targets supported mypy protobuf versions,
allowing us to drop 2.7 as a mypy_protobuf runtime and simplify futher

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -5,30 +5,70 @@ on: [push, pull_request, workflow_dispatch]
 
 jobs:
   run_test:
-    name: Test on python ${{ matrix.python-version }}
+    name: ${{ matrix.py-ver-mypy-protobuf }} / ${{matrix.py-ver-mypy-target}} / ${{matrix.py-ver-unit-tests}} protoc/mypytarget/unittests py versions
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [2.7, 3.5, 3.6, 3.7, 3.8]
+        # Running mypy-protobuf itself
+        # py-ver-mypy-protobuf: []
+        # --python-version=X target passed into mypy
+        # py-ver-mypy-target: []
+        # Run unit tests - validating that correct typechecking usages
+        # of the google.protobuf library in various python versions
+        # also run at runtime
+        # py-ver-unit-tests: []
+
+        # Explicitly list out each part of the matrix we're using, rather
+        # than automatically multiplying the set in together, so we don't
+        # over-test combinations unnecessarily.
+        include:
+          # Include 3.5, 2.7 for the mypy-target
+          - py-ver-mypy-protobuf: 3.8.6
+            py-ver-mypy-target: 3.5
+            py-ver-unit-tests: 3.8.6
+          - py-ver-mypy-protobuf: 3.8.6
+            py-ver-mypy-target: 2.7
+            py-ver-unit-tests: 3.8.6
+
+          # Include py 2.7, 3.6 - 3.8 for mypy-protobuf runtime
+          - py-ver-mypy-protobuf: 3.6.12
+            py-ver-mypy-target: 3.5
+            py-ver-unit-tests: 3.8.6
+          - py-ver-mypy-protobuf: 3.7.9
+            py-ver-mypy-target: 3.5
+            py-ver-unit-tests: 3.8.6
+          - py-ver-mypy-protobuf: 2.7.18
+            py-ver-mypy-target: 3.5
+            py-ver-unit-tests: 3.8.6
+
+          # Include python runtime of google.protobuf in 2.7
+          - py-ver-mypy-protobuf: 3.8.6
+            py-ver-mypy-target: 3.5
+            py-ver-unit-tests: 2.7.18
     steps:
       - name: Install Protoc
         uses: arduino/setup-protoc@v1
         with:
-          version: '3.13.0'
+          version: '3.14.0'
       - uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


run_test.sh
```diff
@@ -32,36 +38,48 @@ find generated -type f -not \( -name "*.expected" -or -name "__init__.py" \) -de
...
-    source $MYPY_VENV/bin/activate
-    if [[ -z $SKIP_CLEAN ]] || [[ ! -e $MYPY_VENV ]]; then
+    source $VENV/bin/activate
+    if [[ -z $SKIP_CLEAN ]] || [[ ! -e $VENV ]]; then
         python3 -m pip install setuptools
         python3 -m pip install git+https://github.com/python/mypy.git@985a20d87eb3a516ff4457041a77026b4c6bd784
     fi
 
     # Run mypy
-    for PY in 2.7 3.5; do
```
nipunn1313, line 50: notably here - a single run of this script only runs a mypy target version (defaulting to 3.5) - however it will update both output.expected if either fails (for convenience).

This should make it easier to selectively compile grpc proto files only on python3.

.github/workflows/main.yml
```diff
@@ -5,30 +5,59 @@ on: [push, pull_request, workflow_dispatch]
...
-    name: Test on python ${{ matrix.python-version }}
+    name: ${{ matrix.py-ver-mypy-protobuf }} / ${{matrix.py-ver-mypy-target}} / ${{matrix.py-ver-unit-tests}} protoc/mypytarget/unittests py versions
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [2.7, 3.5, 3.6, 3.7, 3.8]
+        # Running mypy-protobuf itself
+        py-ver-mypy-protobuf: [3.8.6]
+        # --python-version=X target passed into mypy
+        py-ver-mypy-target: [3.5, 2.7]
```
Evgenus, line 15: I think you dont need this. It is complicated to combine iterative matrix with additional include list. Using just include list would be much clear. (also you have only 2 alternatives accordingly to the matrix)

.github/workflows/main.yml
```diff
@@ -5,30 +5,59 @@ on: [push, pull_request, workflow_dispatch]
...
-      - name: Set up python ${{ matrix.python-version }}
-        uses: actions/setup-python@v2
+      - name: Install pyenv
+        uses: gabrielfalcao/pyenv-action@v7
         with:
-          python-version: ${{ matrix.python-version }}
+          # Ideally - we could specify only the versions we need, but blocked
+          # on https://github.com/gabrielfalcao/pyenv-action/issues/135
+          # versions: ${{matrix.py-ver-mypy-protobuf}}, ${{matrix.py-ver-unit-tests}}, 3.8.6
+          versions: 3.7.9, 3.6.12, 2.7.18
```
Evgenus, line 49: well, you can run `pyenv install` inside run_test

run_test.sh
```diff
@@ -32,36 +38,48 @@ find generated -type f -not \( -name "*.expected" -or -name "__init__.py" \) -de
...
-            for PY in 2.7 3.5; do
-                mypy --custom-typeshed-dir=$CUSTOM_TYPESHED_DIR --python-version=$PY python/mypy_protobuf.py test_negative/ generated/ > test_negative/output.expected.$PY || true
-            done
-            exit 1
-        fi
-    done
+    mypy --custom-typeshed-dir=$CUSTOM_TYPESHED_DIR --python-version=$PY_VER_MYPY_TARGET --pretty --show-error-codes python/mypy_protobuf.py test/ generated/
+    if ! diff <(mypy --custom-typeshed-dir=$CUSTOM_TYPESHED_DIR --python-version=$PY_VER_MYPY_TARGET python/mypy_protobuf.py test_negative/ generated/) test_negative/output.expected.$PY_VER_MYPY_TARGET; then
+        echo -e "${RED}test_negative/output.expected.$PY_VER_MYPY_TARGET didnt match. Copying over for you. Now rerun${NC}"
+        for PY_VER_MYPY_TARGET in 2.7 3.5; do
```
Evgenus, line 63: this loop is not necessary because we need to run tests twice anyway to check both targets

.github/workflows/main.yml
nipunn1313: It was motivated by a desire to increase this to include 3.6, 3.7, 3.8, 3.9 - so it would become the big one soon.

Perhaps more clear would be two separate build plans - one intending to test many targets holding the rest constant.

One intending to test many py-ver-mypy-protobuf holding the mypy-target constant.

That requires more refactoring - I'll punt on it for now.

.github/workflows/main.yml
nipunn1313: The advantage of doing it here - is that we can enable caching of the pyenv root directory - where installation is happening in an earlier stage, so have some preference to pushing installations/dependencies outside of the run_test.sh phase.

I do acknowledge that there are some dependency installations still in run_test.sh, but I'm hoping to move those out of there over time - to separate the "build" and "test" phases.

run_test.sh
nipunn1313: yep - that makes sense, will update. Thanks

PR #210, opened 2021-02-07
Changed stubs to use callable instead of def

### LINKED ISSUE #209: Use UnaryUnaryMultiCallable (and others) instead of function definition in Stubs
Hey! Kudos for releasing V2! It's such a great improvement!

I'm especially fond of the new grpc stubs! they're quite nice.

One thing of note, however is that you are currently using a function definition in service stubs.

However, the  actual implementation is a `MultiCallable`. Using that would allow for passing in additional arguments (like `timeout`), and also access to it's non-calling members (`future` and `with_call`)

See here:

``` Python

# Currenty:
def Generate(self,
    request: global___GenerateRequest,
) -> global___GenerateResponse: ...

# Should Be: 
    Generate : grpc.UnaryUnaryMultiCallable[
        global___GenerateRequest,
        global___GenerateResponse] = ...
```
However, this assumes the use of the grpc-stubs typings (in order for the generics to work). That said, I think it's already a dependency of this module. 

Is there a reason for this decision?

MHDante:
Changed type generation for `grpcio` stubs to use the `MultiCallable` API ([see here](https://grpc.github.io/grpc/python/grpc.html#multi-callable-interfaces)) . This requires using the `grpc-stubs` typings for grpcio. This change should allow calling stub methods with common parameters (`timeout`, `metadata`, etc.) as well as calling methods on the `MultiCallable` object (e.g. `my_stub.MyRpcMethod.future()`).

Fixes #209


MHDante:
Do not merge until 
https://github.com/shabbyrobe/grpc-stubs/pull/10
is merged and 
https://github.com/dropbox/mypy-protobuf/pull/210/files#diff-4d7c51b1efe9043e44439a949dfd92e5827321b34082903477fd04876edb7552R7
is changed back.

MHDante:
Do not merge, waiting on:
https://github.com/shabbyrobe/grpc-stubs/pull/10

After which, this should be changed back:
https://github.com/dropbox/mypy-protobuf/pull/210/files#diff-4d7c51b1efe9043e44439a949dfd92e5827321b34082903477fd04876edb7552R7

nipunn1313:
looks reasonable to me - but disclaimer, I've never actually used this for grpc.

Would appreciate if you ran the tests and added the generated ".expected" files to the PR - so we can review the different generated code. (Can be easier to see the difference in behavior based on the diff on the generated files)

Notably - want to make sure things are backward compatible (or if they aren't - that we put a disclaimer in our release notes)

If it's looking promising, would also appreciate an entry in CHANGELOG (under upcoming). And you'd be welcome to add yourself in the contributors section (under README) - and make any changes that seem reasonable to you to the GRPC doc under README.

MHDante:
Hmm. I've run the tests. but it seems to be [skipping some of the import mangling](https://github.com/dropbox/mypy-protobuf/pull/210/files#diff-8b209c561a349fcb9c6be6cb773a71a317d2cfe64dff1a9b3659df59808e7a3cL5-R7).  

I don't think this is the result of any of the code I've written/modified. It could be that the expected files were not updated on a previous commit. (or some flag I'm not setting in the test script?)

Please let me know, otherwise, I've updated the changelog and expected files as well.

nipunn1313:
oh interesting - I'll look into it! In theory - the CI testsuite should fail if expected files weren't updated on prior commits, but i agree w/ you - that in practice that doesn't seem to have happened.

We did refactor CI quite a bit in January - so I can double check on that stuff. In any case, thank you for the contribution!

nipunn1313:
this looks good to me! I'll give @Evgenus an opportunity to chime in - but I think that since this is passing the tests that he wrote, this is ok with me!

nipunn1313:
One idea - to make sure we don't regress this in the future -
May want to add some usages in test_grpc_usage.py - which illustrate the use cases which you are enabling (eg a usage w/ timeout parameter).

The testsuite should run those tests as well as run mypy over those tests.

CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 ## Upcoming
 
+- Changed type generation for `grpcio` stubs to use the `MultiCallable` API ([see here](https://grpc.github.io/grpc/python/grpc.html#multi-callable-interfaces)) . This requires using the `grpc-stubs` typings for grpcio. This change should allow calling stub methods with common parameters (`timeout`, `metadata`, etc.) as well as calling methods on the `MultiCallable` object (e.g. `my_stub.MyRpcMethod.future()`).
```
Evgenus, line 3: @nipunn1313 are you ok that grpc-stubs will be a strict dependency? BTW, this need to be specified as requirement.

Evgenus: Ah.. I've forgot that it already is a dependency. Ok let's wait until we can use official master.

Evgenus:
> this looks good to me! I'll give @Evgenus an opportunity to chime in - but I think that since this is passing the tests that he wrote, this is ok with me!

This is great. I will take a look on how it works with real code in pycharm. Also I would like to review tests, will do it soon.

Evgenus:
@MHDante Thank you. Great job. I will review shortly.

Evgenus (approved):

MHDante:
I added some tests. LGTM?

nipunn1313:
Nice tests! Awesome stuff!

PR #211, opened 2021-02-18
Ordered enum classes  to work better with Pyright

MHDante:
The order that enums are currently generated leads to an undeclared type being used for both the mangling reference and the field references. This might not be an issue in mypy, but in Pyright, it causes the linter to not be able to recognize the source type for these fields:

image

I've simply switched around the declaration to this: 
image

Outside of giving the members a little more room to breathe, it reduces the declaration order issue.

However, there is still the unknown type in the generic EnumTypeWrapper. I think I will investigate that in the pyright repo (since I assume it works with mypy)


nipunn1313:
This looks great!
Can you confirm you've signed the CLA?

nipunn1313:
Can you add an entry to the CHANGELOG.md as well?

MHDante:
Done so with both.

PR #228, opened 2021-07-07
Support -V and --version flags to protoc-gen-mypy

### LINKED ISSUE #220: Bake version in code during builds
Hello!

When we compile protobufs with protoc, we also generate a version.py file at the top level of each packages. This file currently contains both the package's version being compiled, and the protoc version used for compiling. 

I'd like to do the same with mypy-protobuf and record its version as well.

Essentially, this file would look like this:

    __version__ = "1.2.3"
    __protoc_version__ = "3.17.1"
    __mypy_protobuf_version__ = "2.5"

However right now mypy-protobuf does not expose its version anywhere, so it would be great if it could be added!

nipunn1313:
Fixes #220

### DIFF
--- a/mypy_protobuf/main.py
+++ b/mypy_protobuf/main.py
@@ -25,6 +25,7 @@ from google.protobuf.internal.containers import RepeatedCompositeFieldContainer
 from google.protobuf.internal.well_known_types import WKTBASES
 from . import extensions_pb2
 
+__version__ = "2.5"
 
 # So phabricator doesn't think mypy_protobuf.py is generated
 GENERATED = "@ge" + "nerated"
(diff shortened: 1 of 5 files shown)


tarmath:
Thank you for the quick fix!

mypy_protobuf/main.py
```diff
@@ -828,6 +829,10 @@ def generate_mypy_grpc_stubs(
 def code_generation() -> Generator[
     Tuple[plugin_pb2.CodeGeneratorRequest, plugin_pb2.CodeGeneratorResponse], None, None
 ]:
+    if len(sys.argv) > 1 and sys.argv[1] in ("-V", "--version"):
+        print(__version__)
```
tarmath, line 833: small nitpick, but noteworthy to mention that usually the output for version also prints a program name and not just the version by itself:

```
$ python -V
Python 3.6.8

$ protoc --version
libprotoc 3.14.0

$ pytest -V
pytest 6.2.4

$ black --version
black, version 21.5b1

$ cargo --version
cargo 1.53.0 (4369396ce 2021-04-27)
```

And sometimes, even more information is added, such as copyright, license, etc...

nipunn1313: good idea.

Something like `mypy-protobuf 2.5` might be helpful

PR #234, opened 2021-07-08
Bump mypy version to 0.910

nipunn1313:


test_negative/output.expected.2.7
```diff
@@ -1,3 +1,4 @@
+test/test_generated_mypy.py:198: error: Argument 1 to "Value" of "_EnumTypeWrapper" has incompatible type "unicode"; expected "str"
```
nipunn1313, line 1: Hey @ilevkivskyi  - do you have any idea why this error is appearing, but only when mypy is run under python 2.7?

Simple repro
```
➜  mypy-protobuf git:(bump_mypy_version) mypy --python-version=2.7 test/test_generated_mypy.py `find test/generated -name "*.pyi" | grep -v grpc`
test/test_generated_mypy.py:198: error: Argument 1 to "Value" of "_EnumTypeWrapper" has incompatible type "unicode"; expected "str"
Found 1 error in 1 file (checked 14 source files)
➜  mypy-protobuf git:(bump_mypy_version) mypy --python-version=3.8 test/test_generated_mypy.py `find test/generated -name "*.pyi" | grep -v grpc`
Success: no issues found in 14 source files
```

nipunn1313: bah - I rubber duckied myself. Here's what the stubs in typeshed look like
```
def Value(self, name: Union[str, bytes]) -> _V: ...
```

Under python2, this isn't actually working as intended. Needs to be Union[unicode, str, bytes] or Union[Text, bytes]. I can submit an update to typeshed.

PR #245, opened 2021-07-14
Fix module level extensions

### LINKED ISSUE #244: proto3 Extensions defined at module level produce bad mypy ExtensionDict types
I'm trying the new versions of `mypy-protobuf` 2.6 together with `types-protobuf` 3.17.0 and I'm getting some errors when using extensions.

```proto
syntax = "proto3";

extend google.protobuf.FieldOptions {
  // Give a bit more semantics to a field so that we can programatically
  // identify sensitive fields to handle with care.
  FieldUsage field_usage = 50000;
}

enum FieldUsage {
  // Many fields are not tagged, this is either because they have not been
  // tagged yet or because they define the user's project and as such do not
  // need special care yet. In the future we might add a tag for those if there
  // is a need.
  UNKNOWN_FIELD_USAGE = 0;

  // A field that may uniquely identify a user.
  PERSONAL_IDENTIFIER = 1;

  // A field that is only used for the app's UX or UI: it does not contain
  // actual user data.
  APP_ONLY = 2;

  // A field whose value is provided by the user as a feedback.
  USER_FEEDBACK = 3;

  // A field populated by the app as a result of an algorithm.
  ALGORITHM_RESULT = 4;
}
```

```py
        extensions = field_descriptor.GetOptions().Extensions
        field_usage = extensions[options_pb2.field_usage]
```

Here's what mypy tells me for line 1 and 2 (respectively)
```sh
Need type annotation for "field_usage"
Invalid index type "FieldDescriptor" for "_ExtensionDict[FieldOptions]"; expected type "_ExtensionFieldDescriptor[FieldOptions, <nothing>]"
```

I've dug a bit and it seems that `mypy-protobuf` types `options_pb2.field_usage` as a generic `google.protobuf.descriptor.FieldDescriptor` instead of having a `_ExtensionFieldDescriptor[FieldOptions, options_pb2.FieldUsage>]`.

Does someone else have this issue? Should I change the generated type?

nipunn1313:
Update types-protobuf pin from typeshed

Depends on https://github.com/python/typeshed/pull/5774
Fixes #244


nipunn1313:
Hi @EPronovost - if you're around in the next few days, would love to get your review on this and dependency https://github.com/python/typeshed/pull/5774

This one won't pass tests until https://github.com/python/typeshed/pull/5774 gets in. With modular typeshed, I should be able to land them one after another.

nipunn1313:
Tests now pass - since python/typeshed#5774 landed and types-protobuf updated (automatically) to 3.17.1.
I updated this diff (and documentation) to reference types-protobuf 3.17.1 as the requirement.

EPronovost:
Sorry for not being quick enough... these look good.  Thanks for doing this!

PR #278, opened 2021-08-19
Avoid relative imports in grpc code

### LINKED ISSUE #277: Not getting the type for the grpc response
The type of `response` in the following code is `Unknown`.

```python
statement = exchange_statement_pb2.ExchangeStatement(title="JohnDoe")
with grpc.insecure_channel("alba:50051") as channel:
    stub = alba_pb2_grpc.AlbaStub(channel)
    response = stub.TagStatement(statement)
```

Is that expected?

I'd expect a typed `TagReply` since my proto looks like this

```proto
service Alba {
    rpc TagStatement (ExchangeStatement) returns (TagReply) {}
}

message TagReply {
    repeated string tags = 1;
}
```

nipunn1313:
Fixes #277

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 - Switch from setup.py to pyproject.toml and setup.cfg per https://packaging.python.org/tutorials/packaging-projects/
 - Remove dependency on grpcio-tools. mypy-protobuf doesn't need it to run. It's only needed to run mypy afterward.
+- Avoid relative imports in `_grpc_pb2.pyi` stubs. grpc stubs themselves don't use relative imports, nor `__init__.py` files
 
 ## 2.9
 
--- a/mypy_protobuf/main.py
+++ b/mypy_protobuf/main.py
@@ -147,11 +147,13 @@ class PkgWriter(object):
         descriptors: Descriptors,
         readable_stubs: bool,
         relax_strict_optional_primitives: bool,
+        grpc: bool,
     ) -> None:
         self.fd = fd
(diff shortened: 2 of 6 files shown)


nipunn1313:
Would appreciate a review from either @MHDante or @Evgenus as well as they've contributed majorly to the GRPC stubs here.

My rationale is that this matches more closely to how the `_grpc_pb2.py` work, and obviates need for `__init__.py` files (at least matches more closely w/ how the `.py` works.

Evgenus (approved):

Evgenus:
Looks good but what if generated files are not in root package?

nipunn1313:
I believe that you need to add your generated files directory to your PYTHONPATH or sys.path for things to work (see https://github.com/protocolbuffers/protobuf/issues/881). I think it's just a shortcoming of protobuf. Looks like the task was wont-fixed in 2017. I don't think that use case is well supported unfortunately.

mypy will find imports via this mechanism https://mypy.readthedocs.io/en/stable/running_mypy.html#finding-imports

For example, in our testsuite, the top of `dummy_pb2_grpc.py` looks like this
```
# Generated by the gRPC Python protocol compiler plugin. DO NOT EDIT!
"""Client and server classes corresponding to protobuf-defined services."""
import grpc

from testproto.grpc import dummy_pb2 as testproto_dot_grpc_dot_dummy__pb2
```
^ not using relative import.

With this diff, the `dummy_pb2_grpc.pyi` will match that - and thus work more similarly in cases like the one in #277

PR #281, opened 2021-08-28
Use docstring style comments instead of # comments

### LINKED ISSUE #280: Improve copying of comments
In https://github.com/dropbox/mypy-protobuf/commit/eb1defa98654832732368ee4247009212d337567, comments from the .proto files were copied over into the type stubs.

The comments in the .proto files I'm using are intended to be docstrings, but mypy-protobuf writes them out to the .pyi files using `#` rather than `"""`, and before message classes rather than on the first line of the class body, so they are not treated as docstrings and so don't show up in the documentation view of the IDE I am using (PyCharm 2021.2).

i.e. I would prefer:

```
# My leading detached comments
class Foo(google.protobuf.message.Message):
    """My leading comments

    My trailing comments
    """
    DESCRIPTOR: google.protobuf.descriptor.Descriptor = ...
    ...
```

rather than the current situation of:

```
# My leading detached comments
# My leading comments
# My trailing comments
class Foo(google.protobuf.message.Message):
    DESCRIPTOR: google.protobuf.descriptor.Descriptor = ...
```

This convention of documentation as a docstring on the first line of the class seems to me to be common enough to justify the suggested pattern. Since there is less of a prevailing convention for how to document attributes of a Python class, I think the current situation (a # comment before the field attribute) is okay for field comments.

nipunn1313:
Fixes #280


jaredkhan:
Thanks for looking into this. 

For my use-case of getting these docstrings to show up in PyCharm, moving the comment down into the top of the class body rather than before the class definition is just as important as changing from comments to docstrings.

nipunn1313:
I put another couple of hours of work into it and got it working. I set up pycharm to make sure it was showing up in the IDE. I would really appreciate a second set of eyes. The github diff should show some example files with docstring comments. Take a look @jaredkhan.

Great suggestion.

jaredkhan:
I claimed in the issue that there was no prevailing convention for attribute docstrings but then found this PEP:
https://www.python.org/dev/peps/pep-0258/#attribute-docstrings

which, among other things, says that docstrings for class-level assignments should be placed *after* the assignment. For example:

```
class AClass:

    c = 'class attribute'
    """This is AClass.c's docstring."""
```

This doesn't end up in any `__doc__` attributes but _is_ read by docutils and I can see that, taking PyCharm as an example again, it can show up in IDEs. That would lead me to suggest that we do change the way field comments are done after all, as follows:

- For non-repeated scalar fields, place the docstring comment after the assignment
- For other fields, place the docstring comment within the property method

For avoidance of doubt, I imagine that would look something like:
```
                for idx, field in enumerate(desc.field):
                    if field.name in PYTHON_RESERVED:
                        continue
                    field_scl = scl + [d.DescriptorProto.FIELD_FIELD_NUMBER, idx]
                    self._write_comments(field_scl)
                    if (
                        is_scalar(field)
                        and field.label != d.FieldDescriptorProto.LABEL_REPEATED
                    ):
                        # Scalar non repeated fields are r/w
                        l("{}: {} = ...", field.name, self.python_type(field))
                        self._write_comments(field_scl)
                    else:
                        # r/o Getters for non-scalar fields and scalar-repeated fields
                        l("@property")
                        l(
                            "def {}(self) -> {}:",
                            field.name,
                            self.python_type(field),
                        )
                        with self._indent():
                            self._write_comments(field_scl)
                            l("pass")
```

jaredkhan:
The above comment applies to Enum values too actually. So

```
enum Test {
  FIRST = 0; // this is the first option
  SECOND = 1; // this is the second option
}
```

would generate something like

```
class _TestEnumTypeWrapper(google.protobuf.internal.enum_type_wrapper._EnumTypeWrapper[_Test.V], builtins.type):
    DESCRIPTOR: google.protobuf.descriptor.EnumDescriptor = ...

    FIRST = Test.V(0)
    """this is the first option"""

    SECOND = Test.V(1)
    """this is the second option"""



FIRST = Test.V(0)
"""this is the first option"""

SECOND = Test.V(1)
"""this is the second option"""
```


meaning write_enum_values would become something like:

```
        for i, val in values:
            if val.name in PYTHON_RESERVED:
                continue

            scl = scl_prefix + [i]

            self._write_line(
                "{} = {}({})",
                val.name,
                value_type,
                val.number,
            )
            self._write_comments(scl)
```

maybe adding an additional newline after the comments if readable_stubs is turned on, since otherwise it could be confusing which comments are attached to which attributes

nipunn1313:
Ooh nice find. Let me try it. Now that I have pycharm set up - I can test if it helps there.

ivucica:
Here's what's done by [protoc-docs-plugin](https://github.com/googleapis/protoc-docs-plugin) when a single line comment is provided:

```python
ListQuotesRequest = _reflection.GeneratedProtocolMessageType('ListQuotesRequest', (_message.Message,), {
  'DESCRIPTOR' : _LISTQUOTESREQUEST,
  '__module__' : 'quotes_pb2'
  ,
  '__doc__': """Request passed to the ListQuotes RPC.
  
  Attributes:
      parent:
           “quotesets/$ID” 
      page_token:
           Not supported at this time.
  """,
  # @@protoc_insertion_point(class_scope:some.project.here.proto.ListQuotesRequest)
  })
_sym_db.RegisterMessage(ListQuotesRequest)
```

Note how it generated `attributes` as well: however, only the attributes/arguments/parameters with a docstring are included.

This doesn't get loaded by pyright, which is why I am interested in docs getting support in mypy-protobuf.

Source message:

```protobuf
// Request passed to the ListQuotes RPC.
message ListQuotesRequest {
  string parent = 1; // "quotesets/$ID"
  string filter = 2;
  string order_by = 3;
  int32 page_size = 4;
  string page_token = 5; // Not supported at this time.
}
```

Putting a docstring below the field also helps: hovering over `quotes_pb2.ListQuotesRequest().page_token`  provides the `Not supported at this time.` doc as well.

```python
class ListQuotesRequest(google.protobuf.message.Message):                                                                                      
    """Request passed to the ListQuotes RPC."""                                                                                                
    DESCRIPTOR: google.protobuf.descriptor.Descriptor = ...                                                                                    
    PARENT_FIELD_NUMBER: builtins.int                                                                                                          
    FILTER_FIELD_NUMBER: builtins.int                                                                                                          
    ORDER_BY_FIELD_NUMBER: builtins.int                                                                                                        
    PAGE_SIZE_FIELD_NUMBER: builtins.int                                                                                                       
    PAGE_TOKEN_FIELD_NUMBER: builtins.int                                                                                                      
    parent: typing.Text = ...                                                                                                                  
    """quotesets/$ID"""                                                                                                                        
    filter: typing.Text = ...                                                                                                                  
    order_by: typing.Text = ...                                                                                                                
    page_size: builtins.int = ...                                                                                                              
    page_token: typing.Text = ...                                                                                                              
    """Not supported at this time."""                                                                                                          
```

Putting a docstring under `@property` also works, as long as `...` is gone:

```python
    TOTAL_SIZE_FIELD_NUMBER: builtins.int                                                                                                      
    NEXT_PAGE_TOKEN_FIELD_NUMBER: builtins.int                                                                                                 
    @property                                                                                                                                  
    def quoteset(self) -> global___QuoteSet:
        """An alternative is to return a QuoteSet with the repeated quote field.                                                               
                                                                                                                                               
        Not done at this time because of filtering."""                                                                                         
    @property                                                                                                                                  
    def quote(self) -> google.protobuf.internal.containers.RepeatedCompositeFieldContainer[global___Quote]: ...                                
    total_size: builtins.int = ...                                                                                                             
```

Putting it under `__init__` also helps. (The "Google-style" args docstring doesn't seem to be particularly parsed by pyright or by Emacs's lsp-pyright, so I am not thinking of it much.)

nipunn1313:
Bam it works! It's awesome! Enum values and class attributes.
Check it out

image

ivucica:
Seems to work!

jaredkhan:
Nice one @nipunn1313! Thanks for spending the time on this, this is a nice improvement 👍

Issue #296, opened 2021-09-22
proto file with double quotes at the end of the message field comment is generated with a syntax error


efpato:
message.proto
```proto
syntax = "proto3";

package message;

message Message {
  // Field "One".
  string field1 = 1;
  // Field "Two"
  string field2 = 2;
}
```

message_pb2.pyi
```python
...
class Message(google.protobuf.message.Message):
    DESCRIPTOR: google.protobuf.descriptor.Descriptor = ...
    FIELD1_FIELD_NUMBER: builtins.int
    FIELD2_FIELD_NUMBER: builtins.int
    field1: typing.Text = ...
    """Field "One"."""

    field2: typing.Text = ...
    """Field "Two""""  <-- [mypy error] EOL while scanning string literal
...
```

sodul:
The issue is that the comments from the proto files are entered as string declarations. This is really not a good idea to do that, comments should be comments and have a `#` in front of them.

This is causing a lot of breakage with out own proto files.

sodul:
@nipunn1313  @jaredkhan can you please revert #281. While this comes from a good idea it breaks code generation under different conditions, namely if the comment starts or ends with `"`, and maybe (have not checked) if the comment ends with `\`.

This is a big regression and I am not a big fan of abusing string declarations as comments, since they can cause further unexpected side effects, even though this use is approved by Guido.

jaredkhan:
Good points, thanks for raising both. The docstring functionality added is quite useful and so I think it's worth considering how we can get it working without these issues.

The important things are:
- to make sure the triple quotes aren't interfered with
- to make sure the string itself doesn't contain syntax errors such as invalid escape sequences

I think the following PR gets most of the way to escaping what needs to be escaped without sacrificing the accessibility of the documentation:
https://github.com/dropbox/mypy-protobuf/pull/297/files
Feel free to give that a try and see if it works for you or have a read to see if you can spot other issues.

Overall, it's of course up to @nipunn1313 what approach to take

sodul:
I have to point out that there is a difference between a docstring, which I fully encourage, and a string declaration used as a comment.

For example:
```
class Foo:
    """This is the docstring for Foo."""  # A docstring

    """Comment about bar."""  # anonymous string declaration
    bar = "value for bar"
    """Comment about baz."""  # anonymous string declaration
    baz = "value for baz"
    my_dict = {
        """Comment about Bob:"""  # string declaration, will be joined
        'bob': 'secret for bob',
[...]

PR #297, opened 2021-09-25
Add some fixes and basic testing of docstring escaping

jaredkhan:
Relates to https://github.com/dropbox/mypy-protobuf/issues/296


mypy_protobuf/main.py
```diff
@@ -269,8 +269,24 @@ def _write_comments(self, scl: SourceCodeLocation) -> bool:
...
         if len(lines) == 1:
-            self._write_line('"""{}"""', lines[0])
+            line = lines[0]
+            if line.endswith('"'):
+                # Docstrings are terminated with triple-quotes, so if the documentation itself ends in a quote,
+                # insert some whitespace to separate it from the closing quotes.
+                # This is not necessary with multiline comments
+                # because in that case we always insert a newline before the trailing triple-quotes.
+                line = line + " "
+            self._write_line('"""{}"""', line)
```
sodul, line 289: As per my comment in #296, declaring a comment with `"""` does not make it a docstring, which is a very common misconception.

https://www.python.org/dev/peps/pep-0257/#what-is-a-docstring

If you really want to use anonymous string declarations then I would extract the code here into its own method so you can fully unittest various use cases to ensure they end up doing what you actually expect.

Something that could be relatively safe would be to use `repr()` on the strings you want to use as comments and just use that since they are safely parsable:

```// my comment``` would then become `' my comment'`. And things like

It does mean that your anonymous string declarations now use `'` instead of `"""`, but there is nothing that technically links `"""` to docstring except for the convention in pep-0257 linked above.

My recommendation would be to let comments be comments, as with `#`, since that what they technically are for, and then add a new option for folks that do prefer the anonymous string declarations because their IDE is configured to mask comments (I know some people do).

Honestly I feel we are fixing a user problem for the few (invisible comments due to IDE configuration), with a technical problem (safe string parsing is complicated), where there was no problem for most users (comments just worked).

File: mypy_protobuf/main.py [hunk 2]
```diff
@@ -269,8 +269,24 @@ def _write_comments(self, scl: SourceCodeLocation) -> bool:
         if sci_loc.trailing_comments is not None:
             lines.extend(self._break_text(sci_loc.trailing_comments))
 
+        lines = [
+            # Replace backslashes with double-backslashes
+            # to remove anything that would become an escape sequence in a Python string.
+            # Escape triple-quotes that would otherwise end the docstring early.
+            line
+                .replace("\\", "\\\\")
```
nipunn1313, line 277: I believe we should use raw strings https://docs.python.org/3/reference/lexical_analysis.html#string-and-bytes-literals - rather than trying to do this by hand.

mypy_protobuf/main.py
```diff
@@ -269,8 +269,24 @@ def _write_comments(self, scl: SourceCodeLocation) -> bool:
...
+                .replace('"""', '\\"\\"\\"')
```
nipunn1313, line 278: I actually suspect this will be insufficient. For example, I think if `"""""` is in a comment, this will end up as `\"\"\"""`.

```
>>> '"""""'.replace('"""', '\\"\\"\\"')
'\\"\\"\\"""'
```

I can think of two options:
1)
```
.replace('"', '\"')
```

2) Something like this
```
while '"""' in line:
  line = line.replace('"""', '\"\"\"')
```

Whatever you prefer.

mypy_protobuf/main.py
nipunn1313: repr seems like a good idea.

I do think that the PEP257 describes
```
String literals occurring immediately after a simple assignment at the top level of a module, class, or __init__ method are called "attribute docstrings".
```
which is what the IDE is leaning on here. I don't think it's a configuration issue - but rather this is what the IDE is looking for by the spec.

nipunn1313 (commented):
If it's possible to replace this entire logic w/ a single call to `repr` that would be ideal.
I'm going to merge this as an improvement, but then follow up to try to fix the other issues. Thanks!

mypy_protobuf/main.py
sodul: Let me clarify. I think that adding docstrings is great but it should only be the docstrings. For the other proto comments that are inclined alongside the attributes, they should remain comments and not strings.

nipunn1313: Do you have concrete examples? Eg in https://github.com/dropbox/mypy-protobuf/blob/main/test/generated/testproto/test_pb2.pyi ?

We write PEP257 docstring comments for

- top of the module
- enum class
- message class
- message getter methods
- extension classes
- service methods
- service classes
- grpc methods
- grpc classes

We write PEP257 attribute docstrings for
- enum values
- message scalar fields

I can't find any comments of the variety you're describing.

PR #298, opened 2021-09-27
Fix special chars further

### LINKED ISSUE #296: proto file with double quotes at the end of the message field comment is generated with a syntax error

nipunn1313:
Based on comments from #297 
Added some additional tests and moved to using raw strings

Fixes #296

### DIFF
--- a/mypy_protobuf/main.py
+++ b/mypy_protobuf/main.py
@@ -271,7 +271,7 @@ class PkgWriter(object):
 
         lines = [
             # Escape triple-quotes that would otherwise end the docstring early.
-            line.replace('"""', '\\"\\"\\"')
+            line.replace('"""', r"\"\"\"")
             for line in lines
         ]
         if len(lines) == 1:
--- a/mypy_protobuf/main.py
+++ b/mypy_protobuf/main.py
@@ -284,11 +284,11 @@ class PkgWriter(object):
                 # This is not necessary with multiline comments
                 # because in that case we always insert a newline before the trailing triple-quotes.
                 line = line + " "
-            self._write_line('"""{}"""', line)
+            self._write_line('r"""{}"""', line)
         else:
             for i, line in enumerate(lines):
                 if i == 0:
-                    self._write_line('"""{}', line)
+                    self._write_line('r"""{}', line)
                 else:
                     self._write_line("{}", line)
             self._write_line('"""')
--- a/mypy_protobuf/main.py
+++ b/mypy_protobuf/main.py
@@ -270,15 +270,13 @@ class PkgWriter(object):
             lines.extend(self._break_text(sci_loc.trailing_comments))
 
         lines = [
-            # Replace backslashes with double-backslashes
-            # to remove anything that would become an escape sequence in a Python string.
             # Escape triple-quotes that would otherwise end the docstring early.
-            line.replace("\\", "\\\\").replace('"""', '\\"\\"\\"')
+            line.replace('"""', '\\"\\"\\"')
             for line in lines
         ]
         if len(lines) == 1:
             line = lines[0]
-            if line.endswith('"'):
+            if line.endswith(('"', "\\")):
                 # Docstrings are terminated with triple-quotes, so if the documentation itself ends in a quote,
                 # insert some whitespace to separate it from the closing quotes.
                 # This is not necessary with multiline comments
--- a/test/generated/testproto/comment_special_chars_pb2.pyi
+++ b/test/generated/testproto/comment_special_chars_pb2.pyi
@@ -18,19 +18,19 @@ class Test(google.protobuf.message.Message):
     D_FIELD_NUMBER: builtins.int
     E_FIELD_NUMBER: builtins.int
     a: typing.Text = ...
-    """Ending with " """
+    r"""Ending with " """
 
     b: typing.Text = ...
-    """Ending with "" """
+    r"""Ending with "" """
 
     c: typing.Text = ...
-    """Ending with \"\"\" """
+    r"""Ending with \"\"\" """
 
     d: typing.Text = ...
-    """Ending with \\"""
+    r"""Ending with \\"""
 
     e: typing.Text = ...
-    """Containing bad escape: \\x"""
+    r"""Containing bad escape: \\x"""
 
     def __init__(self,
         *,
--- a/test/generated/testproto/comment_special_chars_pb2.pyi
+++ b/test/generated/testproto/comment_special_chars_pb2.pyi
@@ -21,6 +21,8 @@ class Test(google.protobuf.message.Message):
     G_FIELD_NUMBER: builtins.int
     H_FIELD_NUMBER: builtins.int
     I_FIELD_NUMBER: builtins.int
+    J_FIELD_NUMBER: builtins.int
+    K_FIELD_NUMBER: builtins.int
     a: typing.Text = ...
     r"""Ending with " """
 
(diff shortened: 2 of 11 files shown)


test/generated/testproto/comment_special_chars_pb2.pyi
```diff
@@ -17,20 +17,36 @@ class Test(google.protobuf.message.Message):
...
 
     e: typing.Text = ...
-    """Containing bad escape: \\x"""
+    r"""Containing bad escape: \x"""
+
+    f: typing.Text = ...
+    r"""Containing \"\"\"" quadruple"""
+
+    g: typing.Text = ...
+    r"""Containing \"\"\""" quintuple"""
```
nipunn1313, line 43: this actually seems to work ok!

nipunn1313:
@jaredkhan and @sodul - would appreciate if you took a look at this.

test/generated/testproto/grpc/dummy_pb2_grpc.pyi
```diff
@@ -8,61 +8,61 @@ import testproto.grpc.dummy_pb2
 import typing
 
 class DummyServiceStub:
-    """DummyService"""
+    r"""DummyService"""
```
sodul, line 11: This is an actual docstring, and I think that is helpful and will help with IDEs that show them. 👍

test/generated/testproto/grpc/dummy_pb2_grpc.pyi
```diff
@@ -8,61 +8,61 @@ import testproto.grpc.dummy_pb2
...
     def __init__(self, channel: grpc.Channel) -> None: ...
     UnaryUnary: grpc.UnaryUnaryMultiCallable[
         testproto.grpc.dummy_pb2.DummyRequest,
         testproto.grpc.dummy_pb2.DummyReply] = ...
-    """UnaryUnary"""
+    r"""UnaryUnary"""
```
sodul, line 16: Not a docstring, can we make it use the real comment syntax instead, which we had in the prior release?

test/generated/testproto/grpc/dummy_pb2_grpc.pyi
```diff
@@ -8,61 +8,61 @@ import testproto.grpc.dummy_pb2
...
     StreamStream: grpc.StreamStreamMultiCallable[
         testproto.grpc.dummy_pb2.DummyRequest,
         testproto.grpc.dummy_pb2.DummyReply] = ...
-    """StreamStream"""
+    r"""StreamStream"""
 
 
 class DummyServiceServicer(metaclass=abc.ABCMeta):
-    """DummyService"""
+    r"""DummyService"""
```
sodul, line 35: An actual docstring 👍

test/generated/testproto/grpc/dummy_pb2_grpc.pyi
```diff
@@ -8,61 +8,61 @@ import testproto.grpc.dummy_pb2
...
     @abc.abstractmethod
     def UnaryUnary(self,
         request: testproto.grpc.dummy_pb2.DummyRequest,
         context: grpc.ServicerContext,
     ) -> testproto.grpc.dummy_pb2.DummyReply:
-        """UnaryUnary"""
+        r"""UnaryUnary"""
```
sodul, line 41: Also a docstring to the method, 👍

test/generated/testproto/test3_pb2.pyi
```diff
@@ -117,13 +117,13 @@ class SimpleProto3(google.protobuf.message.Message):
     @property
     def bool(self) -> global___OuterMessage3: ...
     OuterEnum: global___OuterEnum.V = ...
-    """Test having fieldname match messagename"""
+    r"""Test having fieldname match messagename"""
```
sodul, line 120: Not a docstring, it will be 'mostly' ignored by the compile and IDEs, in the best case.

mypy_protobuf/main.py
```diff
@@ -270,25 +270,23 @@ def _write_comments(self, scl: SourceCodeLocation) -> bool:
             lines.extend(self._break_text(sci_loc.trailing_comments))
 
         lines = [
-            # Replace backslashes with double-backslashes
-            # to remove anything that would become an escape sequence in a Python string.
             # Escape triple-quotes that would otherwise end the docstring early.
-            line.replace("\\", "\\\\").replace('"""', '\\"\\"\\"')
+            line.replace('"""', '\\"\\"\\"')
```
sodul, line 274: Use a raw string declaration so you do not need to escape `\`
```suggestion
            line.replace('"""', r'\"\"\"')
```

mypy_protobuf/main.py
```diff
@@ -270,25 +270,23 @@ def _write_comments(self, scl: SourceCodeLocation) -> bool:
...
             for line in lines
         ]
         if len(lines) == 1:
             line = lines[0]
-            if line.endswith('"'):
+            if line.endswith(('"', "\\")):
```
sodul, line 279:
```
if line.endswith(('"', r"\")):
```

test/generated/testproto/comment_special_chars_pb2.pyi
sodul: But these are not docstrings, they are somewhat ignored by the python compiler and IDEs as if they were comments ... as such I think they should remain as comments, and only actual docstrings should be made into strings.

I see that you test for comments like:
```
// this comment has "quotes"
// this comment has """Triple quotes"
```

Have you covered the case where we would have this in the proto files:
```
// "quotes" can be a problem in comments.
// """Triple quotes""" just as well
```

test/generated/testproto/test3_pb2.pyi
jaredkhan: These get processed by docutils (and therefore by some IDEs, PyCharm for example) as per ([PEP-258>Attribute Docstrings](https://www.python.org/dev/peps/pep-0258/#id40) in a way that they wouldn't otherwise if they weren't strings.

nipunn1313: I confirmed before adding these originally that PyCharm does in fact process these per PEP-258 as an Attribute Docstring.

test/generated/testproto/comment_special_chars_pb2.pyi
nipunn1313: These are PEP257 Attribute Docstrings - defined as:

String literals occurring immediately after a simple assignment at the top level of a module, class, or __init__ method are called "attribute docstrings".

nipunn1313: I can add these two tests. I have seen comments like this before.
Good idea to add them as test cases.

You're right that testing these scenarios are unfortunately tricky.
Also totally agree that docstrings are probably a mis-feature. I am a bigger fan of comments being comments - with docutils-esque programs parsing comments rather than relying on strings / `__doc__` attributes (rust does this for example). That being said - my fanhood is not relevant - gotta do what works for the python ecosystem.

sodul: Ah, I was not aware of Attribute Docstrings. You learn something new every day.

nipunn1313: I learned about the distinction from you in fact - when you linked the PEP! Thanks for educating us about it.
I knew I had tested it out with PyCharm - but hadn't dug into the weeds about what it was looking for.

nipunn1313:
I'll add @sodul and @jaredkhan as contributors - and acknowledge you in the readme. Appreciate y'all!

ilianovoselov-foodpairing:
Why did you decide to use raw strings here? Shouldn't they be actually usual strings, so escapes would be handled properly?

ilianovoselov-foodpairing:
@nipunn1313 with raw strings:
```
"""
@generated by mypy-protobuf.  Do not edit manually!
isort:skip_file
"""
import google.protobuf.descriptor
import google.protobuf.message

DESCRIPTOR: google.protobuf.descriptor.FileDescriptor = ...

class TestMessage(google.protobuf.message.Message):
    r"""test \"\"\" "test" """
    DESCRIPTOR: google.protobuf.descriptor.Descriptor = ...
    def __init__(self,
        ) -> None: ...
global___TestMessage = TestMessage

>>> test_pb2.TestMessage.__doc__
'test \\"\\"\\" "test" '
```

ilianovoselov-foodpairing:
With plain string:
```
>>> print(test_pb2.TestMessage.__doc__)
test """ "test" 
```

nipunn1313:
Originally used raw strings to handle backslashes in the original input. However, you're right that it messes up quotation marks. If we substitute backslashes in the original input, all the tests seem to pass - even w/o raw strings, so that's an improvement. That being said - I don't think you can actually use `__doc__` with pyi files? Maybe some IDEs do their own parsing.

ilianovoselov-foodpairing:
Sure they do their own parsing, but I'm assuming their parser should converge to python's reference grammar and this could become issue now or later

PR #310, opened 2021-11-19
Use Optional[T] in constructor for proto3_optional fields

icgood:
As shown below, the constructor does function differently when passed `None` vs an "empty" value for fields declared `optional` in proto3 (HasField specifically). I believe this justifies using `field_name: Optional[Foo] = ...` in the constructor.

```pycon
>>> from test3_pb2 import SimpleProto3
>>>
>>> one = SimpleProto3(an_optional_string=None)
>>> one.an_optional_string
''
>>> one.HasField('an_optional_string')
False
>>>
>>> two = SimpleProto3(an_optional_string='')
>>> two.an_optional_string
''
>>> two.HasField('an_optional_string')
True
```


nipunn1313:
Cool! Thanks for the find!This appears to be a relatively new feature in protobuf which we added support for about a year ago with #157.

Stabilized in february 2021 with 3.15.0
https://github.com/protocolbuffers/protobuf/blob/edc8a3182811cc39272549ff894793e1fff4aaab/CHANGES.txt#L412
Originally added as an experimental feature in 3.12.0 back in May 2020
https://github.com/protocolbuffers/protobuf/blob/master/CHANGES.txt#L672

Happy to take this in! Thanks for the contribution. Few action items

- Add a regression test.

One good place to add one might be here. Add a test `test_constructor_proto3` which constructs with both examples you provided above
https://github.com/dropbox/mypy-protobuf/blob/26c1578925bbbaccbe92b02a7fc05dc7bcf4555c/test/test_generated_mypy.py#L423
I also suspect we'll want to update the HasField test in the same file to cover this case now that it's stabilized (see original tests added in #157). 

- Update CHANGELOG.md with a quick description of this bugfix - linking to https://github.com/protocolbuffers/protobuf/blob/master/docs/field_presence.md#presence-in-proto3-apis

Your diff inspired me to think about what to do about our flag `relax_strict_optional_primitives` - https://github.com/dropbox/mypy-protobuf#relax_strict_optional_primitives

I think we want to keep this flag for the time being, for folks that aren't using `optional` on their fields. It will catch bugs for them. We should update the docs to constrain to proto3 fields without `optional`. We can reference https://github.com/protocolbuffers/protobuf/blob/master/docs/field_presence.md#presence-in-proto3-apis

nipunn1313:
if you're busy or this sounds like a lot - let me know and I can pick it up from you. Contribution is appreciated.

nipunn1313:
Looks great! Thank you!

PR #349, opened 2022-02-12
Update README

miaachan:
This PR updates the comment about PEP 563 on the current use of enum `ValueType`. 

PEP 563 was announced as default in 3.10, but this decision has been rolled back.
fyi: https://mail.python.org/archives/list/python-dev@python.org/thread/CLVXXPQ2T2LQ5MP2Y53VVQFCXYWQJHKZ/


nipunn1313:
Looks great! Thanks for the update.

PR #360, opened 2022-03-17
Use PEP604 syntax for optional

### LINKED ISSUE #358: Use PEP 604 syntax where possible
Typeshed is still not able to able to use PEP 604 syntax _everywhere_ due to several remaining bugs in mypy's implementation. However, these bugs _only_ impact PEP 604 unions in _runtime_ contexts (type aliases and use as class bases). Typeshed now always uses PEP 604 unions in variable annotations and function declarations, which are unaffected by the remaining mypy bugs.

When generating stubs for typeshed, it would be great if mypy-protobuf could also use the new-style syntax in those situations :)

nipunn1313:
Fixes #358


AlexWaygood:
Thanks!

PR #362, opened 2022-03-21
Prefer typing over typing_extensions for python versions that support it.

fergyfresh:
When using this package I noticed the mypy built types `*.pyi` files still relied on `typing_extensions` after we upgraded to 3.10. I know that `Literal` has been in typing since 3.8+ and was wondering if this kind of change was something you'd want.

This worked for my project, but I'm not aware of the long term vision for this project. We could certainly have smarter shims for this too if we'd like, but this was just my initial cut at trying to reduce the `typing_extension` dependency on the auto-generated files.

### DIFF
--- a/mypy_protobuf/main.py
+++ b/mypy_protobuf/main.py
@@ -86,6 +86,11 @@ PROTO_ENUM_RESERVED = {
     "items",
 }
 
+if sys.version_info > (3, 9):
+    typing_package_name = "typing"
+else:
+    typing_package_name = "typing_extensions"
+
 
 def _mangle_global_identifier(name: str) -> str:
     """
@@ -333,13 +338,13 @@ class PkgWriter(object):
 
             l(f"class {enum_helper_class}:")
             with self._indent():
-                l(
+                l(G
                     "ValueType = {}('ValueType', {})",
                     self._import("typing", "NewType"),
                     self._builtin("int"),
                 )
                 # Alias to the classic shorter definition "V"
-                l("V: {} = ValueType", self._import("typing_extensions", "TypeAlias"))
+                l("V: {} = ValueType", self._import(typing_package_name, "TypeAlias"))
             l(
                 "class {}({}[{}], {}):",
                 etw_helper_class,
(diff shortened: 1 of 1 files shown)


nipunn1313:
Hi - unfortunately, we actually need to know the version of python mypy is targeting, not the version that mypy-protobuf itself is running.

For example, some projects generate stub files using mypy-protobuf (and check them in).
Later, they run mypy across their codebase + generated stubs files multiple times, targeting all supported versions of python (eg 3.8, 3.9, 3.10)

Thus, in order for this to work, the conditional needs to be code-generated into the stubs themselves.
We could do that, but I'm not sure it leads to a benefit.

fergyfresh:
First off, thanks for the quick response! I'm unsure what you mean by you're comment though because that's precisely what I thought we can do with this code as I tested generating the stubs from the `*.proto` file locally with two separate python environments, one with 3.6 and one with 3.10 and as we can see from the failing tests the stubs. Unless you mean they want to generate the stubs once and use them across many different python versions.

Rather than complicating the stub generation code would it possible to add an optional flag that you could target which minimum python versions you want to support such that you can reduce the dependency on `typing_extensions`? That's the primary goal of this PR; to not force `typing_extensions` on the project, the only alternative now is to manually fix the generated stubs, right?

I don't think moving the check into the generated stubs themselves is worth it as you've previously pointed out.

nipunn1313:
I believe you're using the same python environment to run mypy-protobuf as well as to run mypy on the generated code. This is not always the case.

For example, you could install python 3.10 and run
```
mypy --python-version 3.6 [/path/to/generated/files]
```
This is not as unusual as it seems at a glance - for projects that want to support multiple versions of python. It's often the case that you generate stubs once but run them through mypy across many target versions.

Some flag like `--min-python-version` could work to help remove the dependency on typing_extensions, but I am hesitant to add additional complexity like this - since it ends up increasing the test matrix for support. Can you provide some background into the motivation here? Typing extensions does not seem unreasonable as a dependency for typed generated code. What's the motivation for really wanting to eliminate it?

fergyfresh:
Its definitely not an unreasonable dependency, the main motivation was that our project no longer needed it and upon digging this was the last piece still importing it and was purely an attempt to try to decouple the dependency. I wasn't sure of the usages that you brought up and I wouldn't want to increase the support burden for a nice to have feature.

nipunn1313:
Felt inspired by this and produced #400.
Gave you credit for the idea and the initial implementation in our README!

Closing this one out.

Issue #378, opened 2022-06-05
Use official style for generated stubs


q0w:
- Style-guide: https://typing.readthedocs.io/en/latest/source/stubs.html
- Test against [flake8-pyi](https://github.com/PyCQA/flake8-pyi) 
- Use the same lint settings like in [python/typeshed](https://github.com/python/typeshed)

e.g.
- `Ellipsis(...)` instead of `pass` for Function body
- str instead of typing.Text (because py2 support was dropped already)
- collections.abc.* instead of typing collections
- remove docstrings

nipunn1313:
(Mostly) Sounds good!

Docstrings are there for a reason - for autocompletion information - since the original _pb2 generated code doesn't generate docstrings - so we'll keep those, but most of these style changes seem reasonable and standardized.

nipunn1313:
Yes - that sounds good in spirit.

I have a preference for generating black/isort conformant stubs and validating them with a testsuite that does `black --check` and `isort --check`. That way black/isort don't become part of the dependencies of mypy-protobuf.

Generating code - and then post-formatting it can make it difficult to understand/debug what the code generator itself is doing, but I think we can achieve the goal without sacrificing this.

Evgenus:
@q0w 

> collections.abc.* instead of typing collections

Where do you see this? I can't find

> remove docstrings

What docstrings? "this code is autogenerated" warnings?

q0w:
@Evgenus check typeshed code and flaks8-pyi rules.

PR #380, opened 2022-06-19
pass -> Ellipsis, typing.Text -> str

Evgenus:
implements #378

### DIFF
--- a/test/generated/testproto/grpc/import_pb2_grpc.pyi
+++ b/test/generated/testproto/grpc/import_pb2_grpc.pyi
@@ -33,7 +33,7 @@ class SimpleServiceServicer(metaclass=abc.ABCMeta):
         context: grpc.ServicerContext,
     ) -> testproto.test_pb2.Simple1:
         """UnaryUnary"""
-        pass
+        ...
 
     @abc.abstractmethod
     def UnaryStream(self,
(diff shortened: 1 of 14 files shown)


test/generated/testproto/nested/nested_pb2.pyi
```diff
@@ -34,7 +34,7 @@ class AnotherNested(google.protobuf.message.Message):
         ONE: AnotherNested._NestedEnum.ValueType  # 1
         TWO: AnotherNested._NestedEnum.ValueType  # 2
     class NestedEnum(_NestedEnum, metaclass=_NestedEnumEnumTypeWrapper):
-        pass
+        ...
```
nipunn1313, line 37: I think to be conformant, this ellipsis would have to go on the same line.

test/generated/testproto/grpc/import_pb2_grpc.pyi
```diff
@@ -33,15 +33,15 @@ class SimpleServiceServicer(metaclass=abc.ABCMeta):
...
+        ...
 
     @abc.abstractmethod
     def UnaryStream(self,
         request: testproto.test_pb2.Simple1,
         context: grpc.ServicerContext,
     ) -> google.protobuf.empty_pb2.Empty:
         """UnaryStream"""
-        pass
+        ...
```
nipunn1313, line 44: I *think* (not 100% sure) to be conformant, this ellipsis isn't necessary because a docstring was emitted.

nipunn1313 (changes requested):
This looks good! Couple of commentsinline .

Can you add a note to CHANGELOG.md about the changes here?

nipunn1313:
Actually, going to merge this - can you follow up with changes to CHANGELOG.md
Trying to do nitpicks on the style is not that helpful until we actually run `black --check` / `isort --check` in CI, so don't worry about those things. Thanks for the improvements!

PR #385, opened 2022-07-20
Save module docstring to top of pyi files

Alphadelta14:
This attempts to save javadoc-like comments as a module docstring to the top of the pyi files.

I love using this project with Sphinx and this was just a piece of missing information that I want to have exposed, especially since I leave tons of documentation about usage at the top of my protobuf files.

I will take a look at tests if this implementation looks okay (I don't want to overstep and rewrite some of the helper functions to be more elegant to this kind of change).


nipunn1313:
If you add some module docstrings to the top of `test.proto` and `test3.proto` that can serve as a test.

PR #547, opened 2023-08-03
Add mypy error code to type ignore.

### LINKED ISSUE #544: Add error code to bare type ignore
Hey, after PR #489 my generated code is failing mypy with option `ignore-without-code` due to bare type ignore comments.

The error I am getting is:
`myfile.pyi:16: error: "type: ignore" comment without error code (consider "type: ignore[misc, type-arg]" instead)  [ignore-without-code]`

Any objections to `[misc, type-arg]` being thrown on the end of the type ignore comments generated here: https://github.com/nipunn1313/mypy-protobuf/blob/main/mypy_protobuf/main.py#L727 ?

Or I can always turn that mypy option off for this particular project.

wwuck:
This supports running mypy with `ignore-without-code`.

Fixes #544

### DIFF
--- a/mypy_protobuf/main.py
+++ b/mypy_protobuf/main.py
@@ -724,7 +724,7 @@ class PkgWriter(object):
         # _ServicerContext is supertyped by grpc.ServicerContext and grpc.aio.ServicerContext
         # So both can be used in the contravariant function parameter position.
         wl(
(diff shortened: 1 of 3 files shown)


nipunn1313:
lgtm thanks!

PR #607, opened 2024-04-24
Mark mangled globals as `TypeAlias`

### LINKED ISSUE #600: `google.protobuf.any_pb2.global__Any` is not marked as `TypeAlias`
https://github.com/nipunn1313/mypy-protobuf/pull/321 added the `TypeAlias` annotation for `class.V`, but using stubs generated by typeshed, there's still a single `Y026` (https://github.com/PyCQA/flake8-pyi/blob/main/ERRORCODES.md / https://docs.astral.sh/ruff/rules/type-alias-without-annotation/) violation in https://github.com/python/typeshed/blob/7d56cd9a6cf6e0a4ea89c68d0397e197aff32cbe/stubs/protobuf/google/protobuf/any_pb2.pyi#L177

Avasam:
Closes #600

- Mark top-level mangled identifiers as `TypeAlias`. (Y026)
- Change the top-level mangling prefix from `global___` to `Global___` to respect Y042 naming convention.

Refs https://github.com/PyCQA/flake8-pyi/blob/main/ERRORCODES.md#list-of-warnings


Avasam:
However I'm unable to run `run_test.sh` (in WSL):

```shell
$ ./run_test.sh
: invalid option
```

Edit: the script was in CRLF...

New issue, it tries to download `https://github.com/protocolbuffers/protobuf/releases/download/v25.3\r/protoc-25.3\r-linux-x86_64.zip` on my machine.

nipunn1313:
mine works. What's the trouble you're seeing?

Avasam:
Just some initial setup stuff I had to work through. Like having my files in CRLF (because I cloned from Windows with autocrlf on... oops) messing up the script, then figuring out how to install and use pyenv. But I think I got it working now.

Avasam:
Ah, if it ends end raising Y042 instead, I don't know if that's much better. Maybe a little bit since they don't seem to exist at runtime
Speaking of which, what *are* they exactly, the `global___`s ? It seems they can only be used for typing in TYPE_CHECKING context?

Avasam:
And now unused type aliases are (rightfully) detected by Y047 ^^"
I feel like that's a nice-to-fix for another time, I'll reintroduce the stubtest entry and simply rename `global___` to `Global___`

nipunn1313:
Yep this all seems reasonable! Prefixing with `__` also sounds great instead of `Global__`, but can save that for another day. Thanks for the contribution!

I'll try to tie up a few more fixes/updates and then release.

PR #611, opened 2024-05-08
Async Stub Annotations

artificial-aidan:
* Generate a set of TypeVars and a generic class
* TypeVars have defaults to match expected behavior
* Overload init methods to get expected types back
* Create AsyncStub type alias

This was a fun one, I think I've maintained existing behavior, while made it better to use. Initializing with an async channel will now get you the correct type without casting.

I found an issue with the negative tests, there seems to be some ambiguity about the note that mypy is returning, so it intermittently fails


artificial-aidan:
See https://github.com/nipunn1313/mypy-protobuf/issues/536 for more discussion

artificial-aidan:
As I always say, the answer is always in github issues

https://github.com/python/mypy/issues/16363

I added `--no-incremental` to the mypy flags when testing for now.

artificial-aidan:
@nipunn1313 bumping this back to the top.

test/generated/testproto/grpc/dummy_pb2_grpc.pyi
```diff
@@ -7,70 +7,159 @@ import abc
...
 
 class _ServicerContext(grpc.ServicerContext, grpc.aio.ServicerContext):  # type: ignore[misc, type-arg]
     ...
 
-class DummyServiceStub:
-    """DummyService"""
-
-    def __init__(self, channel: typing.Union[grpc.Channel, grpc.aio.Channel]) -> None: ...
-    UnaryUnary: grpc.UnaryUnaryMultiCallable[
+_MTVDummyService0 = typing_extensions.TypeVar(
```
nipunn1313, line 26: don't love the name here. Is there a way to make the naming based on the service name / method name - rather than these iterating numbers?

Perhaps instead of {i} use the {method.name}

`MTV` also feels rather opaque. "MethodTypeVar" isn't particularly useful to the enduser (who will probably see these types when hovering around in vscode).

Perhaps we don't need the "_MTV" prefix at all?

test/test_grpc_async_usage.py
```diff
@@ -53,7 +53,7 @@ async def test_grpc() -> None:
     server = make_server()
     await server.start()
     async with grpc.aio.insecure_channel(ADDRESS) as channel:
-        client: dummy_pb2_grpc.DummyServiceAsyncStub = dummy_pb2_grpc.DummyServiceStub(channel)  # type: ignore
+        client = dummy_pb2_grpc.DummyServiceStub(channel)
```
nipunn1313, line 56: nice!

nipunn1313:
sorry for the delay - I hadn't been focusing on this for a while (I'm unlikely to be highly responsive, but this very-long-wait isn't expected in general - aplogies).

The PR now requires a rebase.

On the whole this looks - great. Some requests.
Could you update CHANGELOG.md in the upcoming section? Would also love to have you add yourself to the contributors section of README.md in the PR as well if you are ok with it.

test/generated/testproto/grpc/dummy_pb2_grpc.pyi
artificial-aidan: It's been a while so I'm going to have to remember why I did the naming like that. I know I wasn't a huge fan of it, but wanted to get something down on paper. Let me see what it looks like to not have the prefix.

artificial-aidan: Ok here we go. I had to keep the leading underscore because of stubtest (left a comment with the SO explaining it, couldn't actually find it in stubtest docs, didn't look too long). So now it is just `_ServiceNameMethodNameType` which shouldn't conflict. Also cleaned it up so the typevar name building so it is more maintainable.

artificial-aidan:
> sorry for the delay - I hadn't been focusing on this for a while (I'm unlikely to be highly responsive, but this very-long-wait isn't expected in general - aplogies).
> 
> The PR now requires a rebase.
> 
> On the whole this looks - great. Some requests. Could you update CHANGELOG.md in the upcoming section? Would also love to have you add yourself to the contributors section of README.md in the PR as well if you are ok with it.

Rebased, let me know if you want me to squash or clean up any before you merge.

nipunn1313:
Ok let's do it!
I am probably going to wait a little bit before releasing again (try to get one more fix in)

tmickel:
Appreciate this one @nipunn1313 and @artificial-aidan! Looks promising for good grpc Python types

artificial-aidan:
@nipunn1313 bump for a release?

artificial-aidan:
@nipunn1313 gentle reminder about getting this released. Not able to bring it into our pipeline yet.

GrahamDennis:
@nipunn1313 bump for release please?

aidandj:
@inickt I am pushing a test release to test.pypi right now. Should have something on main soon.

PR #631, opened 2024-09-05
Implement a new parameter to generate only async stubs

### LINKED ISSUE #536: A better way to annotation the AsyncStub
I am writing async code.

The proto file is like these:
```proto
service Continuum{
  rpc Tick(TickRequest) returns (HistoryResult);
}
```

The pyi file is partially like these:
```python
class ContinuumStub:
    def __init__(self, channel: typing.Union[grpc.Channel, grpc.aio.Channel]) -> None: ...
    Tick: grpc.UnaryUnaryMultiCallable[
        essence.service_pb2.TickRequest,
        essence.service_pb2.HistoryResult,
    ]

class ContinuumAsyncStub:
    Tick: grpc.aio.UnaryUnaryMultiCallable[
        essence.service_pb2.TickRequest,
        essence.service_pb2.HistoryResult,
    ]
```

The ContinuumStub has 2 different params for its __init__ function, which are Channel and aio.Channel. When using aio.Channel, the Stub instance need a type:ignore or typing.cast to convert to AsyncStub, like:
```python
self._channel =grpc.aio.insecure_channel(
            target=self.target,
        )
stub: service_grpc.ContinuumAsyncStub = service_grpc.ContinuumStub(
    self._channel
)  # type:ignore
```
typing.cast is not working here because there isn't a true AsyncStub here, AsyncStub only exists in pyi file.

To solve this problem, maybe using Generic or overload the Tick function in pyi file? 
Any idea?

llucax:
This commit adds a new parameter `async_only` to the plugin, that when present, will only generate the `async` stubs.

The current `{}AsyncStub` classes are *fake*, they are not present in the .py file, so they can only be used to write type hints. Using them also only works if the user manually casts the sync stub to the `async` stub.

Since most users will only want to use either the sync or the `async` stub, this new parameter will allow generating only the stubs that are suited for the user's use case, and allow them to have less hacky code.

Fixes #536.


llucax:
A draft until I get some confirmation that something like this would be accepted. Then I can continue updating the docs and tests (please let me know if anything should be updated).

artificial-aidan:
I think it would make more sense for this to be an enum option. Default to both, async only, sync only

alex-pirogov:
great feature, would like to have it implemented

aidandj:
Superseded by #694

PR #683, opened 2025-10-22
Aidan/pep 702 support

aidandj:
- Support [PEP702](https://peps.python.org/pep-0702/) deprecations
  - Message deprecations are supported
  - Field deprecations are not. This may be possible with init overloads
  - Service deprecations are supported for Sync stubs
    - Not for async stubs
  - Enum message deprecation is supported
    - Enum field deprecation is not

Note: this diff will be much simpler after the previous python version PR (or something similar) is merged. Just the 2nd commit can be reviewed.

### DIFF
--- a/mypy_protobuf/main.py
+++ b/mypy_protobuf/main.py
@@ -2,6 +2,7 @@
 """Protoc Plugin to generate mypy stubs."""
 from __future__ import annotations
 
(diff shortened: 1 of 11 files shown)


nipunn1313:
exciting! Looks good.

PR #710, opened 2025-12-08
Overload __new__

aidandj:
Mypy can't handle a large number of generic overloads (https://github.com/python/mypy/issues/19622) so instead generate 2 different stub classes and override __new__. Seems to have the same functionality, without the generics overhead. Minor feature change in the revealed type of a stub class.


nipunn1313 (approved):
sure - this works.

aidandj:
Although, the error is just in the type checking of the pyi stub, it could technically just be ignored I believe.

subnix:
> This would be technically correct, but it would also be a breaking change, so I think I'm leaning towards the technically incorrect, but non breaking change that is up there now.

Sounds reasonable to me.

subnix (approved):

llucax:
I used this before, and I will continue using it, we have a xkcd situation here:

image

This new overriding of `__new__` instead of `__init__` is *breaking* `griffe` (which in turn *breaks* [mkdocstring](https://mkdocstrings.github.io/python/)). It triggers a warning, but in `mkdocs` strict mode, it effectively makes rendering the documentation fail.

I guess there is no easy fix here, but rising it up in case anyone can see a viable solution to this.

PR #723, opened 2026-04-07
Add typing for Has/ClearField and WhichOneof when they aren't available

aidandj:
On messages without these defined, the typing falls through to the generic version of these methods, which then errors.

```
ValueError: Field Message.field1 does not have presence.
```

This adds `typing.Never` to arguments for functions that do not exist.


nipunn1313 (approved):
very cool idea!

PR #724, opened 2026-04-13
Deprecate if file deprecation option is set

aidandj:
If the deprecated file option is set, deprecate everything inside of it.


nipunn1313 (approved):
cool. I wasn't familiar with this option until now.

PR #729, opened 2026-08-07
fix: suppress generated descriptor override warning

### LINKED ISSUE #721: DESCRIPTOR incompatibility with types-protobuf
In pyright with the latest version of mypy-protobuf (5.0) and types-protobuf, I receive the following error:

```
error: "DESCRIPTOR" overrides symbol of same name in class "Message"
    Variable is mutable so its type is invariant
      Override type "Descriptor" is not the same as base type "google.protobuf.descriptor.Descriptor | google._upb._message.Descriptor" (reportIncompatibleVariableOverride)
```

I noticed in https://github.com/python/typeshed/commit/2b2ea7185bc953bb232597c0a5e1bf7fc988a042 that the DESCRIPTOR type of Message was updated to also include `_upb_Descriptor`.

Would it make sense to update the type of the generated DESCRIPTOR field in the same way?

Neallin-917:
Fixes #721

Generated message DESCRIPTOR declarations keep the public google.protobuf.descriptor.Descriptor type and suppress only Pyright reportIncompatibleVariableOverride. This avoids changing downstream typing behavior or raising the types-protobuf version floor.

The checked-in stubs were regenerated from the updated generator.

Validation:
- PY_VER_MYPY_PROTOBUF=3.14.0 PY_VER_MYPY=3.14.0 PY_VER_UNIT_TESTS=3.14.0 PYTHON_PROTOBUF_VERSION=6.32.1 ./run_test.sh (30 passed)
- Pyright 1.1.407 on a generated stub outside this repository configuration (0 errors)
- black 24.3.0 --check
- isort 6.1.0 --check


aidandj (approved):
I think we can start with this. If needed we can roll it back into an option if needed. I don't see comments breaking anyones workflow though.

